PR #412, opened 2014-10-13
Convert to click

sudorandom:
This is very much a work in progress. There are still a lot of modules to be converted and a lot of testing that needs to be added and adjusted to work again.

That said, this should actually work for the modules that are implemented. Feel free to test it out and report bugs.

Additional Changes:
- Adds verbose option that can be specified multiple times. I think this should eventually replace --debug.
- Add --fixture option which will use the testing fixture client instead of an actual SoftLayer API client. This can and should be used for demoing and testing. Be warned that not all fixtures will be complete enough to use with all CLI commands.
- CLI modules are loaded using entry-points. This will allow for 3rd party and internal-only extensions to the CLI.

### DIFF
--- a/SoftLayer/CLI/core.py
+++ b/SoftLayer/CLI/core.py
@@ -145,7 +145,6 @@ def cli(ctx,
         config=None,
         debug=0,
         verbose=0,
-        timings=False,
         proxy=None,
         really=False,
         fixtures=False,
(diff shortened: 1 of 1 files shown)


sudorandom:
Thanks coveralls. I appreciate the timeliness of this update.

PR #429, opened 2014-10-29
Issues57 - groundwork for monitoring 

allmightyspiff:
This is just some of the groundwork for a new set of methods to manage and view monitoring information.  
Currently "sl monitor status" will just show you the SERVICE_PING results for each server, and the last date it was checked.

screen shot 2014-10-29 at 2 59 35 pm

I wanted to get this pull request in early incase anyone has any architecture concerns.


underscorephil:
ping @sudorandom

SoftLayer/managers/monitor.py
```diff
@@ -0,0 +1,85 @@
...
+    Monitoring Manager/helpers
+
+    :license: MIT, see LICENSE for more details.
+"""
+
+
+from SoftLayer import utils
+
+
+class MonitoringManager(utils.IdentifierMixin, object):
```
sudorandom: `utils.IdentifierMixin` isn't needed. You don't specify ways to look up whatever a "monitor" would be based on something other than IDs.

SoftLayer/managers/monitor.py
```diff
@@ -0,0 +1,85 @@
...
+                'fullyQualifiedDomainName',
+                'primaryBackendIpAddress',
+                'primaryIpAddress',
+                'datacenter',
+                'networkMonitors[lastResult,queryType]'
+            ]
+
+            kwargs['mask'] = ('[mask[%s]]'
+                              % (','.join(vs_items)))
+        function = getattr(self.account, call)
```
sudorandom: This works well without having to use getattr(): self.account.call(call, **kwargs)

SoftLayer/managers/monitor.py
```diff
@@ -0,0 +1,85 @@
...
+
+class MonitoringManager(utils.IdentifierMixin, object):
+    """Manages monitoring for Hardware and Virtual Servers
+
+    :param SoftLayer.API.CLient client: an API client instance
+    :param string server_tpe: should be either
+                  'Hardware_Server' or 'Virtual_Guest'
+    """
+
+    def __init__(self, client, server_type='Hardware_Server'):
```
sudorandom: I really don't like the `server_type` context... This makes one instance of the manager not re-usable between hardware and virtual guests... So this, in affect, is a HardwareMonitoringManager or VirtualGuestHardwareManager depending on what the value of server_type is. It's context, and most other managers have no additional context other than the API client.

Regarding how removing this would affect get_status(), this is a case where it probably should be broken into get_status_for_virtual_server() and get_status_for_hardware(), because that's what you're really doing (taking into the server_type context).

SoftLayer/managers/monitor.py
```diff
@@ -0,0 +1,85 @@
...
+         1 = Warning
+         2 = OK
+         >2 = something strange happened
+        """
+
+        mask = "lastResult, subnet[virtualGuests,hardware]"
+        agent = self.server.getNetworkMonitors(id=server_id, mask=mask)
+        return agent
+
+    def list_hardware_status(self, **kwargs):
```
sudorandom: I can live with how list_hardware_status() and list_guest_status() work currently, but not get_status() and `__init__()`

SoftLayer/managers/monitor.py
allmightyspiff: A good point. My thought at the moment is to have Hardware/VirtualGuestMonitorManager extend MonitoringManager and have them call the parent init. And then move out the list_X_status to each manager. I think that might be a bit more elegant all while trying to avoid duplicating code as much as possible.

SoftLayer/CLI/monitor/status.py
```diff
@@ -0,0 +1,82 @@
...
+    results = hardware + guest
+    for server_object in results:
+        server = utils.NestedDict(server_object)
+        for monitor in server['networkMonitors']:
+            try:
+                res = monitor['lastResult']['responseStatus']
+                date = monitor['lastResult']['finishTime']
+                ip_address = monitor['ipAddress']
+                monitor_type = monitor['queryType']['name']
+            except KeyError:
```
sudorandom: To be more explicit, maybe it's better to do this instead:

```
if 'lastResult' in monitor:
 ...
else:
...
```

The way it works now, a key error for another value could happen and subsequently hidden.

PR #457, opened 2014-11-24
Simplified server provisioning

### LINKED ISSUE #417: Hardware ordering with v4
There are now 3 different methods for ordering hardware (4, if you count Hardware_Server::createObject). They are enumerated below:

* **Normal Hardware Ordering** - uses packages to define different chassis which have different CPU and memory options. Currently, prices inside the package are used to determine which numerical option (ex. 4 CPU cores) matches which price object. The package type is "BARE_METAL_CPU".
* **Bare metal cloud (bmc)** - A limited set of options that was made to (ideally) be more reliable in terms of stock. Hardware_Server::createObject and Hardware_Server::getCreateObjectOptions were made to support bare metal cloud. However, in order to share as much of the existing hardware ordering code that already existed, those interfaces are not used for the CLI/Manager implementation. There are conditionals for BMC servers for ordering, option lookup and canceling. The package type is "BARE_METAL_CORE".
* **Fast hardware provisioning** - Like bare metal cloud, has a limited set of options. However, this is done in a different way than BMC with the concept of order "presets". The suite of presets has its own package (200) and its own package type (BARE_METAL_CPU_FAST_PROVISION). This is currently **not** implemented in the CLI. Here are some API calls that I've found useful:
```python
client['Product_Package'].getActivePresets(id=200)
client['Product_Package_Preset'].getLowestPresetServerPrice(id=68)  # 68 is one of the presets from the previous call.
```

This issue is to track the discussion with the goal of defining what hardware ordering should look like for the version 4 CLI. My primary question is: **Should we support all three ways of hardware ordering in the CLI and in the managers?** Would it be sufficient to only support the fast provisioning servers? Will the fast server provisioning options be implemented with createObject and getCreateObjectOptions? If we're going to support the normal way of ordering hardware, can the manager interface take more of the burden in regard to looking up price ids? Because of the way the interface is, it's impossible to improve upon the implementation without breaking things both in the python and API side.

sudorandom:
Nearly complete. Only TODO is to document arguments in the place_order doc block.

This will resolve #417.


SoftLayer/CLI/server/create.py
```diff
@@ -6,63 +6,25 @@
...
 @click.option('--postinstall', '-i', help="Post-install script to download")
 @helpers.multi_option('--key', '-k',
                       help="SSH keys to add to the root user")
-@helpers.multi_option('--disk', help="Disk sizes")
-@click.option('--controller', help="The RAID configuration for the server")
-@click.option('--private',
-              is_flag=True,
-              help="Forces the VS to only have access the private network")
-@click.option('--network', '-n', help="Network port speed in Mbps")
-@click.option('--template', '-t',
```
sudorandom: This should be added back. I think it was removed by mistake.

sudorandom:
@underscorephil, @allmightyspiff Let me know your thoughts on this.

PR #472, opened 2014-12-16
Adding return value to create_iscsi function

mah5057:
The function 'create_iscsi()' does not return the SoftLayer_Container_Product_Order_Receipt object, this information is needed for applications that need to know the exact volume that was just provisioned in the same scope as the provisioning call. The resourceId of the iscsi volume can then be found by looking up the billing item by the SoftLayer_Billing_Order_Item id in the 'SoftLayer_Billing_Order_Item' service.


underscorephil:
Thanks for the PR!

LGTM.

@sudorandom should we add this in for the 4.0 release?

PR #478, opened 2015-01-21
 add placeorder and verify order for SoftLayer_Billing_Order_Quote

Neetuj:


sudorandom:
@underscorephil is there any difference between using Billing_Order_Quote::placeOrder and Product_Order::placeOrder? If the answer is no then this PR isn't needed at all.

sudorandom:
Closing until this is confirmed to be needed.

PR #479, opened 2015-01-22
Add argument timezone to set MAINTENANCE_WINDOW in SoftLayer/manager/vs....

ryosuke:
...py.

I want upgrade vs's memory from 1024MB to 2048MB, type "sl vs upgrade ID --memory 2048".
Then maintenance time is not immediate. Because not set timezone to datetime.now() in vs.py.
Add import pytz package for use timezone object.

### DIFF
--- a/SoftLayer/managers/vs.py
+++ b/SoftLayer/managers/vs.py
@@ -9,6 +9,7 @@ import datetime
 import itertools
 import socket
 import time
+import pytz
 
 from SoftLayer.managers import ordering
 from SoftLayer import utils
@@ -636,7 +637,7 @@ class VSManager(utils.IdentifierMixin, object):
                            'Upgrade',
             'prices': prices,
             'properties': [{'name': 'MAINTENANCE_WINDOW',
-                            'value': str(datetime.datetime.now())}],
+                            'value': str(datetime.datetime.now(tz=pytz.timezone("posixrules")))}],
             'virtualGuests': [{'id': int(instance_id)}],
         }
         if prices:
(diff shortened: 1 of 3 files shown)


sudorandom:
I'm having an issue with this particular fix. "posixrules" isn't recognized on my system. I think this might be closer to a more general solution:
```
datetime.datetime.utcnow(pytz.utc).strftime("%Y-%m-%d %H:%M:%S%z")
```
This is likely an issue to begin with because the timezone information isn't displayed with doing `str(datetime.datime.now())`. Even using `strftime()` with `datetime.datetime.utcnow()` won't display the timezone info since it assumes the default timezone IS UTC. I am guessing that on the API side the assumption is that the default timezone is CST.

Also, pytz should be added as a requirement in setup.py and tools/requirements.txt. We might be able to get rid of the requirement by implementing a [datetime.tzinfo](https://docs.python.org/2/library/datetime.html#datetime.tzinfo) class to use as the timezone instead of `pytz.utc`. There's actually an example for a UTC tzinfo class in the docs (second code block after the anchor): https://docs.python.org/2/library/datetime.html#datetime.tzinfo.fromutc

ryosuke:
Hi, sudorandom. Thanks for your message.
I think I need sent to API timezone when we will exec "sl vs upgrade ....".
I will sent you pull request next patch soon. Next patch use UTC, but use pytz for replace timezone. sorry :(.

Regard.

sudorandom:
I'm closing this in favor of not introducing a dependency (as detailed in #493).

PR #489, opened 2015-02-17
Adds coveralls to travis build file

sudorandom:
Coveralls should make it make it painfully obvious whenever the coverage percentage goes down.


sudorandom:
Well that's not right....

PR #494, opened 2015-02-19
Adds `sl setup` command

### LINKED ISSUE #459: Better naming of `sl config setup`
It's a little hard to remember exactly how to invoke `sl config setup`. It should probably be aliased to something easier, like just `sl config` or `sl setup`.

sudorandom:
Now you just have to remember `sl setup` to set up the command line client. Resolve #459.

### DIFF
--- a/SoftLayer/CLI/routes.py
+++ b/SoftLayer/CLI/routes.py
@@ -47,6 +47,7 @@ ALL_ROUTES = [
     ('config', 'SoftLayer.CLI.config'),
     ('config:setup', 'SoftLayer.CLI.config.setup:cli'),
     ('config:show', 'SoftLayer.CLI.config.show:cli'),
+    ('setup', 'SoftLayer.CLI.config.setup:cli'),
 
     ('dns', 'SoftLayer.CLI.dns'),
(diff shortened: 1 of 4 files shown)


SoftLayer/CLI/routes.py
```diff
@@ -47,6 +47,7 @@
     ('config', 'SoftLayer.CLI.config'),
     ('config:setup', 'SoftLayer.CLI.config.setup:cli'),
     ('config:show', 'SoftLayer.CLI.config.show:cli'),
```
Neetuj: since "sl conflict setup" and "sl setup" do the same thing .. either we should deprecate one (whichever one sounds less intuitive) or we should document that these 2 commands do the same thing ..else it can leave the users confused

sudorandom: I'd normally agree, but this is THE command that has to be run before using any other part of the CLI. It should be front and center. That's the motivation for putting it in the main CLI listing. The old location is also valid because that's properly namespaced. It involves the configuration of the CLI so it's under 'config'. There are only two subcommands there: show and config. I would be confused if I didn't see a way to edit the configuration there.

allmightyspiff:
+1, I always forget how to setup the client :)

PR #497, opened 2015-03-02
Fix server cost reporting

sergiocarlosmorales:
Update server manager and CLI to properly show the recurring cost of
the server

### DIFF
--- a/SoftLayer/CLI/server/detail.py
+++ b/SoftLayer/CLI/server/detail.py
@@ -80,7 +80,9 @@ def cli(env, identifier, passwords, price):
 
     if price:
         table.add_row(['price rate',
-                       result['billingItem']['nextInvoiceTotalRecurringAmount']])
+                       utils.lookup(result,
+                                    'billingItem',
+                                    'nextInvoiceTotalRecurringAmount')])
 
     if passwords:
         pass_table = formatting.Table(['username', 'password'])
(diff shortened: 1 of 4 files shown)


sergiocarlosmorales:
The failed Travis check looks like a flake, could you guys confirm?

The reason for this change is because the recurringFee property off the billing item only considers the base cost of the server; it doesnt properly report associated costs like OS fee (if any), RAM (like if it was upgraded) etc.

SoftLayer/CLI/server/detail.py
```diff
@@ -80,7 +80,7 @@ def cli(env, identifier, passwords, price):
 
     if price:
         table.add_row(['price rate',
-                       result['billingItem']['recurringFee']])
+                       result['billingItem']['nextInvoiceTotalRecurringAmount']])
```
sudorandom: flake8 is complaining about this line since it's over 80 characters long.

I recommend doing this:
```python
        table.add_row(['price rate',
                       utils.lookup(result,
                                    'billingItem',
                                    'nextInvoiceTotalRecurringAmount')])
```

sudorandom:
BTW, the lower health from landscape wasn't caused by the code in this PR. It was caused by landscape adding a new check.

SoftLayer/CLI/server/detail.py
sergiocarlosmorales: duh, missed that one. fixed.

sudorandom:
LGTM

PR #452, opened 2014-11-13
Rename all the things

sudorandom:
* the base SoftLayer module is renamed to softlayer (SoftLayer -> softlayer)
* the SoftLayer.CLI module is renamed to softlayer.cli (SoftLayer.CLI -> SoftLayer.cli)
* tests have been moved up to the base directory (SoftLayer.tests -> tests)

Issue #425, opened 2014-10-28
Rename sl Command


sudorandom:
The debian package has a patch which renames `sl` to `SL`. This is because a very important program is already using `sl` [0]. Therefore, we should probably use a different name.

Here are my suggestions in order of my preference.

 * slcli
 * sl-cli
 * sl-cmd
 * softlayer

The old name will probably hang around just to inform users to use the new command.

[0] - http://www.cyberciti.biz/tips/displays-animations-when-accidentally-you-type-sl-instead-of-ls.html

allmightyspiff:
+1 slcli

sudorandom:
Resolved via #504

PR #504, opened 2015-03-12
Use slcli command over sl

sudorandom:
#452 is a big change and it has been some time and several other large changes have been merged in since the great refactor. Since the PR stalled for so long, the merging effort is about equal to simply re-doing the changes. Therefore, this PR just implements the renaming of the CLI from `sl` to `slcli`.

I'll close #452 and re-implement it, but it might not be in time for the v4 release.

### DIFF
--- a/SoftLayer/CLI/deprecated.py
+++ b/SoftLayer/CLI/deprecated.py
@@ -1,0 +1,16 @@
+"""
+    SoftLayer.CLI.deprecated
+    ~~~~~~~~~~~~~~~~~~~~~~~~
+    Handles usage of the deprecated command name, 'sl'.
+    :license: MIT, see LICENSE for more details.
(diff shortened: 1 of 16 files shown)


SoftLayer/CLI/deprecated.py
```diff
@@ -0,0 +1,17 @@
...
+    ~~~~~~~~~~~~~~~~~~~~~~~~
+    Handles usage of the deprecated command name, 'sl'.
+    :license: MIT, see LICENSE for more details.
+"""
+import sys
+
+import click
+
+
+@click.command()
```
sudorandom: I think making this a click command is a mistake since it will error out before showing the message in most cases.

sudorandom:
Resolves #425

Issue #506, opened 2015-03-19
The SoftLayer_Ticket::addAttachedFile method fails when tried to attach images


MiguelHigorre:
The SoftLayer_Ticket::addAttachedFile method fails when tried to attach images, the following message is raised
SoftLayerAPIError: SoftLayerAPIError(SoftLayer_Ticket): No valid authentication headers found.

Requirements:
Python 2.6 or 2.7
SoftLayer API python client (3.3.1) 

It seems to be a problem related to Python 2 since using Python 3 or higher the SoftLayer_Ticket::addAttachedFile method works.

The problem is reproduced with the following code

------------------------------------
import SoftLayer
import pprint

apiUsername = 'set-me'
api_key = 'set-me'

apiUsername = 'set-me'
api_key = 'set-me'
ticket_id = set-me
client = SoftLayer.Client(username= apiUsername,api_key=api_key)

f = open('/path/to/image.png', 'rb')
filecontents = f.read()
attachedFile = { 'data': filecontents, 'filename': 'image.png'}
result = client['Ticket'].addAttachedFile(attachedFile, id=ticket_id)
pprint.pprint(result)

-------------------------------------

Here is the traceback for the Python issue.

Traceback (most recent call last):
  File "addAttachedFile.py", line 24, in <module>
    result = client['Ticket'].addAttachedFile(attachedFile, id=13410275)
  File "/usr/local/lib/python2.7/dist-packages/SoftLayer/API.py", line 351, in call_handler
    return self(name, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/SoftLayer/API.py", line 319, in call
    return self.client.call(self.name, name, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/SoftLayer/API.py", line 184, in call
    return transports.make_xml_rpc_api_call(request)
  File "/usr/local/lib/python2.7/dist-packages/SoftLayer/transports.py", line 137, in make_xml_rpc_api_call
    ex.faultCode, ex.faultString)
SoftLayer.exceptions.SoftLayerAPIError: SoftLayerAPIError(SoftLayer_Ticket): No valid authentication headers found.

gordonjm:
I'very been investigating this issue and have seen that the authentication header is constructed, and is part of the payload delivered with the request. However I do not yet understand why this exception is thrown when a Python 2 interpreter is used.

gordonjm:
[...]

PR #507, opened 2015-03-20
The SoftLayer_Ticket::addAttachedFile method fails when tried to attach images #506

gordonjm:
A proposed fix for the "No authentication headers found" problem in issue #506.

### DIFF
--- a/SoftLayer/transports.py
+++ b/SoftLayer/transports.py
@@ -117,10 +117,10 @@ class XmlRpcTransport(object):
             # In Python 2 xmlrpc_client.dumps(...) does not
             # automatically perform base64 encoding.
             if interpreter_version < 3:
-                for dictionary in largs:
+                for index, dictionary in enumerate(largs):
                     if 'data' in dictionary:
-                        data = largs[1]['data']
-                        largs[1]['data'] = base64.b64encode(data)
+                        data = dictionary['data']
+                        largs[index]['data'] = base64.b64encode(data)
                         break
 
             url = '/'.join([request.endpoint, request.service])
--- a/SoftLayer/transports.py
+++ b/SoftLayer/transports.py
@@ -8,12 +8,16 @@
 from SoftLayer import exceptions
 from SoftLayer import utils
 
+import base64
+
 import importlib
 import json
 import logging
 import time
 
 import requests
+
+import sys
 
 LOGGER = logging.getLogger(__name__)
 # transports.Request does have a lot of instance attributes. :(
@@ -107,6 +111,17 @@ class XmlRpcTransport(object):
                 }
 
             largs.insert(0, {'headers': headers})
+
+            interpreter_version = sys.version_info[0]
+
+            # In Python 2 xmlrpc_client.dumps(...) does not
+            # automatically perform base64 encoding.
+            if interpreter_version < 3:
+                for dictionary in largs:
+                    if 'data' in dictionary:
+                        data = largs[1]['data']
+                        largs[1]['data'] = base64.b64encode(data)
+                        break
 
             url = '/'.join([request.endpoint, request.service])
             payload = utils.xmlrpc_client.dumps(tuple(largs),
--- a/SoftLayer/transports.py
+++ b/SoftLayer/transports.py
@@ -112,11 +112,9 @@ class XmlRpcTransport(object):
 
             largs.insert(0, {'headers': headers})
 
-            interpreter_version = sys.version_info[0]
-
             # In Python 2 xmlrpc_client.dumps(...) does not
             # automatically perform base64 encoding.
-            if interpreter_version < 3:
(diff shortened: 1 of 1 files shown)


SoftLayer/transports.py
```diff
@@ -108,6 +112,17 @@ def __call__(self, request):
 
             largs.insert(0, {'headers': headers})
 
+            interpreter_version = sys.version_info[0]
+
+            # In Python 2 xmlrpc_client.dumps(...) does not
+            # automatically perform base64 encoding.
```
Neetuj: to tackle too many local variable restriction of landscape .. you could may be direclty compare sys.version_info[0] <3  

not sure who/why the limit of 15 local variables  is/was enforced

gordonjm:
I eliminated "interpreter_version" in favor of "sys.version_info[0]" to get to the local variables limit.

sudorandom:
Are we sure that it works this way with Python 3? From a library standpoint with no knowledge of the server's API endpoints I don't know how it would ever assume that an arbitrary parameter needs to be base64 encoded without using a user specifying the xmlrpc_client.Binary type or making assumptions (which is what this PR is making). Any other endpoint that takes an object that has a "data" key is assumed to be base64, which has the following problems:

 * Objects with a 'data' attribute may have a type other than base64
 * Attributes of base64 aren't always named 'data'

To me, it seems like the user should give the base64 argument the base64ed value. This should work today with no code change to the library required and fits more in line with SLDN documentation. When it says it expects a base64 encoded value, you should give it a base64 encoded value. Example:
```python
attachedFile = { 'data': base64.b64encode(filecontents), 'filename': 'image.png'}
result = client['Ticket'].addAttachedFile(attachedFile, id=ticket_id)
```

or potentially use a [xmlrpc.Binary](https://docs.python.org/2/library/xmlrpclib.html#binary-objects) value:
```python
attachedFile = { 'data': xmlrpc.Binary(filecontents), 'filename': 'image.png'}
result = client['Ticket'].addAttachedFile(attachedFile, id=ticket_id)
```

or as a SoftLayer-client type (so that other transports like REST will be able to support this):
```python
attachedFile = { 'data': softlayer.Base64(filecontents), 'filename': 'image.png'}
result = client['Ticket'].addAttachedFile(attachedFile, id=ticket_id)
```

gordonjm:
In Python 3 the image file is encoded as base64. In Python 2 it is not. 
This can be verified by dumping the results to standard output using the 
client in the associated issue. One thing I noticed with explicit base64 
encoding (passing filecontents to base64.b64encode) is that Python 3 
appears to encode the already encoded data before sending the request. 
However all is well in Python 2 and 3 when filecontents is passed to 
xmlrpc.Binary. Given the good points you've made I feel that this pull 
request should be killed. The way forward appears to be for the end-user 
to explicitly perform base64 encoding, or form a binary object via 
xmlrpc.Binary.

On 03/24/2015 08:44 PM, Kevin McDonald wrote:
>
> Are we sure that it works this way with Python 3? From a library 
> standpoint with no knowledge of the server's API endpoints I don't 
> know how it would ever assume that an arbitrary parameter needs to be 
> base64 encoded without using a user specifying the 
> xmlrpc_client.Binary type or making assumptions (which is what this PR 
> is making). Any other endpoint that takes an object that has a "data" 
> key is assumed to be base64, which has the following problems:
>
>   * Objects with a 'data' attribute may have a type other than base64
>   * Attributes of base64 aren't always named 'data'
>
> To me, it seems like the user should give the base64 argument the 
> base64ed value. This should work today with no code change to the 
> library required and fits more in line with SLDN documentation. When 
> it says it expects a base64 encoded value, you should give it a base64 
> encoded value. Example:
>
> attachedFile=  {'data': base64.b64encode(filecontents),'filename':'image.png'}
> result=  client['Ticket'].addAttachedFile(attachedFile,id=ticket_id)
>
> or potentially use a xmlrpc.Binary 
> <https://docs.python.org/2/library/xmlrpclib.html#binary-objects> value:
>
> attachedFile=  {'data': xmlrpc.Binary(filecontents),'filename':'image.png'}
> result=  client['Ticket'].addAttachedFile(attachedFile,id=ticket_id)
>
> or as a SoftLayer-client type (so that other transports like REST will 
> be able to support this):
>
> attachedFile=  {'data': softlayer.Base64(filecontents),'filename':'image.png'}
> result=  client['Ticket'].addAttachedFile(attachedFile,id=ticket_id)
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/softlayer/softlayer-python/pull/507#issuecomment-85779506>.
>

-- 

Joseph Gordon
Software Engineer, Business Applications Development
SoftLayer, an IBM Company
4849 Alpha Rd Dallas, TX 75244
214.873.8216x38216 office | 469.394.8816 mobile | jgordon@softlayer.com

PR #510, opened 2015-04-01
Removes billing until it gets much more polish

sudorandom:


briancline:
Added #511 to make sure this gets added back and improved after v4.0.0.

Issue #514, opened 2015-04-03
Solidify Client Construction


sudorandom:
I want to make the following changes:

* Rename SoftLayer.API.Client to SoftLayer.API.BaseClient
* Remove all other arguments from SoftLayer.API.BaseClient except for:
  * transport
  * auth
* Transports will now be constructed with endpoint_url, timeout, proxy and user_agent
* a new create_client_from_env() function will take all current arguments as SoftLayer.API.Client's __init__ arguments and behave the same. It will also have an alias to SoftLayer.API.Client in order to maintain backwards compatibility. Using SoftLayer.API.Client will be deprecated.

This accomplishes the following goals:
 * Generally, removes magic from initialization of an API client
 * Allows for the construction of an API client without possibly loading environmental and file-based configuration options. The fact that this isn't possible has lead to initializing of a client only to set the auth handler to None since it was automatically set by reading config file.
 * Makes transport-specific defaults for timeouts, endpoints, etc happen in an intuitive place.

PR #516, opened 2015-04-07
Improves REST transport + client fixup

sudorandom:
Fully implements the REST transport to support masks, filters, limit, offset, auth. Also implements #514, which should make building the API client a bit less confusing long-term.


sudorandom:
I'm going to abandon the REST transport. Some of the semantics are the same and the motivation behind improving it has since dissipated.

sudorandom:
I will lift the other client changes, however.

PR #518, opened 2015-04-10
Implements Client Improvements 

### LINKED ISSUE #514: Solidify Client Construction

sudorandom:
This is the continuation of #516. It removes most of the REST transport additions. This should resolve #514.

### DIFF
--- a/SoftLayer/API.py
+++ b/SoftLayer/API.py
@@ -12,11 +12,20 @@ from SoftLayer import config
 from SoftLayer import consts
 from SoftLayer import transports
 
+# pylint: disable=invalid-name
+
+
 API_PUBLIC_ENDPOINT = consts.API_PUBLIC_ENDPOINT
 API_PRIVATE_ENDPOINT = consts.API_PRIVATE_ENDPOINT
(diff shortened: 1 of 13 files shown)


sudorandom:
I'm totally going to abuse my power and merge this. I want these changes to land before v4 since a couple specifics would warrant another major version bump.

PR #524, opened 2015-04-14
Hardware cancel consolidation

### LINKED ISSUE #522: problem with hourly bare metal servers immediate cancelation in CLI and manager
slcli server cancel --immediate <server id> does not produce an error, but cancels server on its monthly anniversary. The cause of the problem is "old" logic in cancel_hardware method of harware.py. In SL API 4.0 harware.py place_order now only supports the fast server provisioning package (see version 4 notes: #469), but cancel_hardware is still looking for a 'bareMetalInstanceFlag'  and using Billing_Item cancelService. It should use Billing_Item cancelItem instead

sudorandom:
This should resolve #522.

* HardwareManager.cancel_metal() is removed in favor of HardwareManager.cancel_hardware()
* HardwareManager.cancel_hardware() now uses SoftLayer_Billing_Item::cancelItem() which (I think) works on most cancelable items.

### DIFF
--- a/SoftLayer/CLI/server/detail.py
+++ b/SoftLayer/CLI/server/detail.py
@@ -35,7 +35,7 @@ def cli(env, identifier, passwords, price):
     result = utils.NestedDict(result)
 
     table.add_row(['id', result['id']])
-    table.add_row(['guid', result['globalIdentifier']] or formatting.blank())
+    table.add_row(['guid', result['globalIdentifier'] or formatting.blank()])
     table.add_row(['hostname', result['hostname']])
     table.add_row(['domain', result['domain']])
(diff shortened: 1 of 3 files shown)


sudorandom:
@underscorephil Do you know if SoftLayer_Billing_Item::cancelItem() is ~the~ method to use that can cancel most things? I know they did a lot of work to consolidate cancellation but it's not obvious in the documentation which one should be used.

underscorephil:
@sudorandom doing some research to make sure my info is up to date. Will keep you in the loop

sudorandom:
@underscorephil so it looks like `SoftLayer_Billing_Item::cancelItem` is the universal cancel method. That will likely simplify a lot. Either way, I think this change is ready.

underscorephil:
LGTM

PR #534, opened 2015-04-26
Adds XML-RPC proxy server to be used for tests

sudorandom:
This allows the bindings to be used without a special transport in order to run tests.

* Setting compression off would still send compression headers. That is now fixed.
* Fixes minor testing errors.

### DIFF
--- a/SoftLayer/API.py
+++ b/SoftLayer/API.py
@@ -192,11 +192,12 @@ class BaseClient(object):
         if self._prefix and not service.startswith(self._prefix):
             service = self._prefix + service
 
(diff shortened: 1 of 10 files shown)


sudorandom:
With this change, a new HTTP server is stood up and torn down for each test case class. This is a real XML-RPC server which responds with mocked and/or fixtured responses. It uses the built xmlrpc parsing library for Python, so it's not to be used as a real server (https://docs.python.org/2/library/xml.html#xml-vulnerabilities) but it's sufficient for testing purposes.

PR #540, opened 2015-05-02
Switch capturing of CDROMs

kbild:
If images of Windows server are created the -all disks flag will lead to an error because CDROM can't be imaged. Added skip for mountType equals CD


sudorandom:
I'm closing this in favor of #544. It still has your change, but with a couple adjustments to the style and tests.

PR #551, opened 2015-05-14
Issue550

Neetuj:
This will correct the argument issue( which is causing the name confusion) 
From UI perspective I think it should be ip rather than ip_address ( too long for an argument name to type) 

I also catch the case when ip_record is null before referencing it. Should i throw an error if its not found ?

### DIFF
--- a/SoftLayer/CLI/loadbal/service_add.py
+++ b/SoftLayer/CLI/loadbal/service_add.py
@@ -24,11 +24,11 @@ import click
 @click.option('--healthcheck-type',
               required=True,
               help="The health check type")
-@click.option('--ip-address',
+@click.option('--ip',
               required=True,
               help="The IP of the service")
 @environment.pass_env
-def cli(env, identifier, enabled, port, weight, healthcheck_type, ip_address):
+def cli(env, identifier, enabled, port, weight, healthcheck_type, ip):
     """Adds a new load balancer service."""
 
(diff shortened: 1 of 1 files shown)


sudorandom:
@Neetuj yeah, it is weird that an ip address is required but it looks like it tries to go on without finding a ip record anyway.

SoftLayer/CLI/loadbal/service_add.py
```diff
@@ -37,10 +37,11 @@ def cli(env, identifier, enabled, port, weight, healthcheck_type, ip_address):
 
     # check if the IP is valid
     ip_address_id = None
-    if ip_address:
+    if ip:
         ip_service = env.client['Network_Subnet_IpAddress']
-        ip_record = ip_service.getByIpAddress(ip_address)
-        ip_address_id = ip_record['id']
+        ip_record = ip_service.getByIpAddress(ip)
+        if len(ip_record)>0:
```
sudorandom: There should be a space before and after the > sign.

Neetuj: Ahh ... Thnx

sudorandom:
The tests have been fixed in master. Please rebase in order to get the changes:

```
git pull --rebase upsteam master
```

sudorandom:
Closed in favor of #562

PR #553, opened 2015-05-14
Resolve Test Woes

sudorandom:


sudorandom:
The fact that travis is failing is super confusing. The initial issue (python3 defaulting to ascii encoding) should have never happened because the proper LANG and LC_ALL environmental variables are set to use en_US.UTF-8 (http://docs.travis-ci.com/user/ci-environment/#Environment-variables).

I am still massively confused by this issue.

sudorandom:
The LANG environment variable was not being passed from tox into the test program (nose). The latest change passes the variable along if it exists and defaults to C.UTF-8 if it doesn't exit. Hopefully that works well for travis and most others.

It turns out that LANG and LC_ALL are very important.

Issue #572, opened 2015-07-07
Exception during performing of the 'slcli vs resume' operation.


1v1a3x:
Everything is Ok during performing of the following operation:
```$ slcli vs pause MyHostIdentifier```

But when I try to resume the virtual host with the following command 
I got an exeption:
```$ slcli vs resume MyHostIdentifier```

```An unexpected error has occured:
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\softlayer-4.0.2-py2.7.egg\SoftLayer\CLI\core.py", line 181, in main
    cli.main()
  File "C:\Python27\lib\site-packages\click-4.0-py2.7.egg\click\core.py", line 644, in main
    rv = self.invoke(ctx)
  File "C:\Python27\lib\site-packages\click-4.0-py2.7.egg\click\core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Python27\lib\site-packages\click-4.0-py2.7.egg\click\core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Python27\lib\site-packages\click-4.0-py2.7.egg\click\core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Python27\lib\site-packages\click-4.0-py2.7.egg\click\core.py", line 464, in invoke
    return callback(*args, **kwargs)
TypeError: resume() takes exactly 2 arguments (1 given)```

1v1a3x:
Verified - Fixed in 
slcli (SoftLayer Command-line), version 4.0.4
commit dca89bdfdaf106e729a0f346fd0ffa36cb5ea2d1

Thanks guys (Y)

PR #573, opened 2015-07-07
Fixes resume command

sudorandom:
This should resolve issue #572.


SoftLayer/CLI/virt/power.py
```diff
@@ -101,6 +101,7 @@ def pause(env, identifier):
 
 @click.command()
 @click.argument('identifier')
+@environment.pass_env
```
Neetuj: BTW I think the same bug has come up multiple times ..

PR #575, opened 2015-07-08
Allow editing of hardware server tags

lalinsky:


Neetuj:
adding a test case for adding, deleting  in vs_tests.py might be useful ..rest LGTM

PR #585, opened 2015-07-23
Fixes several incorrect descriptions

sudorandom:


sudorandom:
This change also improves the test_load_all test. It now actually runs each command with the `--help` option.

Issue #593, opened 2015-07-31
KeyError getting server detail from slcli


ajohnson1:
SoftLayer-4.0.4

c:\Python27>scripts\slcli server detail 240292
An unexpected error has occured:
Traceback (most recent call last):
  File "C:\Python27\lib\site-packages\SoftLayer\CLI\core.py", line 181, in main
    cli.main()
  File "C:\Python27\lib\site-packages\click\core.py", line 644, in main
    rv = self.invoke(ctx)
  File "C:\Python27\lib\site-packages\click\core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Python27\lib\site-packages\click\core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Python27\lib\site-packages\click\core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Python27\lib\site-packages\click\core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "C:\Python27\lib\site-packages\click\decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "C:\Python27\lib\site-packages\click\core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "C:\Python27\lib\site-packages\SoftLayer\CLI\server\detail.py", line 100, in cli
    tag_row.append(tag['tag']['name'])
KeyError: 'tag'

Feel free to report this error as it is likely a bug:
    https://github.com/softlayer/softlayer-python/issues

This problem occurs on finding details of some bare metal machines, but not all.

sudorandom:
This should actually be fixed in the master branch and will go out in the next release. Thanks for bringing up the issue!

PR #594, opened 2015-07-31
 make sure if key does not exist no error is thrown

Neetuj:
use get instead of ['key'] ..should solve #593


sudorandom:
This already fixed in master (which is why this is conflicting)

PR #596, opened 2015-08-03
Adds Shell / Moves fixtures

sudorandom:
This commit adds a shell that can be started by simply calling 'slcli' with no command. The shell can be exited using ctrl + d. The shell has autocomplete. The shell probably needs a way to set global options (--fixtures, --debug, --proxy, --config, --timings, etc). In the future, this mode can be used to prompt for user input.

Fixtures were moved to a higher level to avoid needing test dependencies (nosetests, testtools, etc) to use fixtures.

This is still a bit of a WiP.


sudorandom:
Example Usage:

```bash
slcli shell
(0)> vs list
....
(0)>
```

Currently, the global options passed in when using `slcli shell` are maintained in every command in the shell. Those options can be overridden on a per-command basis. For example

```bash
slcli --fixtures shell
(0)> vs list
:.....:..........:..............:.............:............:........:
:  id : hostname :  primary_ip  :  backend_ip : datacenter : action :
:.....:..........:..............:.............:............:........:
: 100 : vs-test1 : 172.16.240.2 : 10.45.19.37 :   TEST00   :   -    :
: 104 : vs-test2 : 172.16.240.7 : 10.45.19.35 :   TEST00   :   -    :
:.....:..........:..............:.............:............:........:
(0)> --no-fixtures vs list
... (non-fixture data)...
```

The only place where this doesn't work as expected is the verbosity/debug level.

SoftLayer/shell/core.py
```diff
@@ -0,0 +1,158 @@
...
+@click.command()
+@environment.pass_env
+@click.pass_context
+def cli(ctx, env):
+    """Enters a shell for slcli."""
+
+    # Set up prompt_toolkit settings
+    app_path = click.get_app_dir('softlayer')
+    if not os.path.exists(app_path):
+        os.makedirs(os.path.dirname(app_path))
```
allmightyspiff: This should be
```python
    os.makedirs(app_path)
```

On mac I get the following. 
**os.path.dirname: /Users/christopher/Library/Application Support**
**app_path: /Users/christopher/Library/Application Support/softlayer**
```bash
~/Code/softlayer-python  $ ./slcli shell
Traceback (most recent call last):
  File "./slcli", line 7, in <module>
    load_entry_point('SoftLayer', 'console_scripts', 'slcli')()
  File "/Users/christopher/Code/softlayer-python/SoftLayer/CLI/core.py", line 178, in main
    cli.main(**kwargs)
  File "/Library/Python/2.7/site-packages/click-3.3-py2.7.egg/click/core.py", line 590, in main
    rv = self.invoke(ctx)
  File "/Library/Python/2.7/site-packages/click-3.3-py2.7.egg/click/core.py", line 936, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Python/2.7/site-packages/click-3.3-py2.7.egg/click/core.py", line 782, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Library/Python/2.7/site-packages/click-3.3-py2.7.egg/click/core.py", line 416, in invoke
    return callback(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/click-3.3-py2.7.egg/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/Library/Python/2.7/site-packages/click-3.3-py2.7.egg/click/core.py", line 416, in invoke
    return callback(*args, **kwargs)
  File "/Users/christopher/Code/softlayer-python/SoftLayer/shell/core.py", line 54, in cli
    os.makedirs(os.path.dirname(app_path))
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/Users/christopher/Library/Application Support'
```

sudorandom: Should be fixed now, thanks!

allmightyspiff:
Spot checked my favorite slcli calls and they all work great.

PR #599, opened 2015-08-09
correct ha argument in firwall add

### LINKED ISSUE #598: Adding Firewall got an unexpected keyword argument 'ha'
Hi there

When trying to add a firewall to a VS instance I got the following exception

C:\Python34\Scripts>slcli firewall add --firewall-type vs 11*****
An unexpected error has occured:
Traceback (most recent call last):
  File "C:\Python34\lib\site-packages\SoftLayer\CLI\core.py", line 181, in main
    cli.main()
  File "C:\Python34\lib\site-packages\click\core.py", line 644, in main
    rv = self.invoke(ctx)
  File "C:\Python34\lib\site-packages\click\core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Python34\lib\site-packages\click\core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Python34\lib\site-packages\click\core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Python34\lib\site-packages\click\core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "C:\Python34\lib\site-packages\click\decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "C:\Python34\lib\site-packages\click\core.py", line 464, in invoke
    return callback(*args, **kwargs)
TypeError: cli() got an unexpected keyword argument 'ha'

Neetuj:
fixes #598

### DIFF
--- a/SoftLayer/CLI/firewall/add.py
+++ b/SoftLayer/CLI/firewall/add.py
@@ -15,7 +15,7 @@ import click
               type=click.Choice(['vs', 'vlan', 'server']),
               help='Firewall type',
               required=True)
-@click.option('--high-availability', '--ha',
+@click.option('--high-availability', '-ha',
               is_flag=True,
               help='High available firewall option')
 @environment.pass_env


SoftLayer/CLI/firewall/add.py
```diff
@@ -15,7 +15,7 @@
               type=click.Choice(['vs', 'vlan', 'server']),
               help='Firewall type',
               required=True)
-@click.option('--high-availability', '--ha',
+@click.option('--high-availability', '-ha',
```
sudorandom: Instead of changing the options, you can just switch the order of "--high-availability" and "--ha". The non-flag is used as the keyword argument. Flags aren't usually more than one character long because you can use them like this: `command -a -b -c`and this: `command -abc`

Neetuj: shouldn't it be  '--high-availability', '-h',  in that case ?

sudorandom: `--ha` is only two characters longer and holds much more information about what it means than just 'h'. Plus, just swapping the two doesn't present an interface change so if any scripts did exist which rely on `--ha` then they would not need to change. For those reasons I prefer just swapping the two arguments (`@click.option('--ha', '--high-availability',`) and not changing the outward interface in any way.

Issue #143, opened 2013-08-01
Wizard-based actions


beittenc:
In addition to allowing the large number of flags and arguments, it would be helpful if the CLI also provided a wizard-based interface for performing various actions. This will decrease the learning curve even further and allow a larger audience base to take advantage of the CLI.

PR #604, opened 2015-08-10
Adds prompts for virtual and hardware server creation.

sudorandom:
* If a user does not pass in an option that is required for `slcli vs create` or `slcli server create` then the user will be prompted for the missing required options. They will not be given the available options (yet). I haven't figured out a good way to do that.
 * Environmental variables prefixed with "SLAPI_" can now be used to default CLI options. For example, this will default to always using raw formatting:

```bash
export SLCLI_FORMAT=raw
slcli vs list
```

 * Fixes the --like option for creating virtual servers.
 * Changes how --like and --template options work to use more of click's feature-set. This was required in order to detect missing options properly.


sudorandom:
Resolves #143.

PR #606, opened 2015-08-10
Changes VSManager.wait_for_ready() to behave more intuitively

### LINKED ISSUE #587: wait_for_ready time limit documentation misleading
In the docs for `wait_for_ready`, the limit parameter is described as:

> :param int limit: The maximum amount of time to wait.

But, if you read the code, limit is actually just the number of times the program sleeps for `delay` before giving up.

`wait_for_ready` with limit=10, delay = 10 waits 10 times as long as limit=10, delay=1.

sudorandom:
* when calling wait_for_ready() it will wait until the given time period and no more (except maybe with the additionally time of an API call)
 * changes `slcli vs create` with the --wait option to give a non-zero exit code if the provision isn't finished in the specified time.
 * Uses a mask to only pull in the relevant information

Resolves #587

### DIFF
--- a/SoftLayer/CLI/virt/create.py
+++ b/SoftLayer/CLI/virt/create.py
@@ -270,9 +270,11 @@ def cli(env, **args):
         output.append(table)
 
         if args.get('wait'):
-            ready = vsi.wait_for_ready(
-                result['id'], int(args.get('wait') or 1))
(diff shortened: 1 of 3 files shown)


sudorandom:
This also changes `slcli vs create` with the `--wait` option to give a non-zero exit code if the provision isn't finished in the specified time.

sudorandom:
@underscorephil rebased

PR #609, opened 2015-08-17
Fix for issue#582. 

ggrunin:
Fixed slcli server create options: added no_public option to allow ordering servers with private only uplink, removed --vlan-public, --vlan-private options which are not applicable for the standard configuration bare metal servers.

### DIFF
--- a/SoftLayer/CLI/server/create.py
+++ b/SoftLayer/CLI/server/create.py
@@ -42,14 +42,9 @@ import click
 @click.option('--postinstall', '-i', help="Post-install script to download")
 @helpers.multi_option('--key', '-k',
                       help="SSH keys to add to the root user")
-@click.option('--vlan-public',
-              help="The ID of the public VLAN on which you want the virtual "
-              "server placed",
-              type=click.INT)
-@click.option('--vlan-private',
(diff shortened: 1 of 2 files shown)


SoftLayer/CLI/server/create.py
```diff
@@ -42,14 +42,9 @@
...
                       help="SSH keys to add to the root user")
-@click.option('--vlan-public',
-              help="The ID of the public VLAN on which you want the virtual "
-              "server placed",
-              type=click.INT)
-@click.option('--vlan-private',
-              help="The ID of the private VLAN on which you want the virtual "
-                   "server placed",
-              type=click.INT)
+@click.option('--no_public',
```
sudorandom: use `--no-public` instead. Click automatically converts dashes to underscores. It's conventional to only have dashes for option names. Also, there are still references to vlan_public and vlan_private in the code below this.

I actually started cleaning this up before your second commit. I can take over this bit from here if you don't mind.

ggrunin: Thanks - please take it over!

sudorandom:
Closing in favor of #611

PR #610, opened 2015-08-17
Remove use of unicode_literals import

sudorandom:


sudorandom:
@underscorephil rebased.

sudorandom:
This causes a warning emitted from click when using the shell.

Issue #547, opened 2015-05-11
server list important columns missing in 4.0


jravetch:
After upgrading to version 4, I lost the ability to see the owner/cpu/memory of each server when using list.  Owner is very important, as it allows me to see quickly what users provisioned which machines.  It also also allows me to see which servers are in the Spare Pool.  Cpu and memory columns allow me to quickly see the cpu/memory allocated of each server.  If you can please bring back these colums ASAP, I would greatly appreciate it.  Moreover, would it be possible to add a spare pool filter so that I can only list servers that are in the spare pool?

sudorandom:
@jravetch The owner column was removed due to performance concerns. The way that data has to be masked in adds 2 full seconds to a typical virtual server listing. Additionally, the amount of horizontal space that we have now is very limited even considering a 140 terminal size. That's why CPU and memory were cut. I can see argument for those being more important than having the ip addresses. The command was modeled after the portal UI, but there's obviously different concerns and limitations for a CLI than a web UI.

There's an issue which should allow you to show all the columns that you want while maintaining performance and UX for a typical case: https://github.com/softlayer/softlayer-python/pull/503. I intend to help out with that when I have the time.

Until then, you can either downgrade to version 3.3.1.
```
pip uninstall softlayer
pip install softlayer==3.3.1
``` 

or you can use the new call-api command to mask in exactly the data that you want. Unfortunately, the data is fairly nested (which probably explains the performance issues), so the best way I've found to show this data is by using the json format:
```bash
slcli --format=json call-api Account getVirtualGuests --mask="id,maxCpu,maxMemory,hostname,billingItem[id,orderItem[id,order[id,userRecord[id,username]]]]"
```

but the normal table version might work better for you if your screen is big enough to avoid line wrapping:
```
slcli call-api Account getVirtualGuests --mask="id,maxCpu,maxMemory,hostname,billingItem[id,orderItem[id,order[id,userRecord[id,username]]]]"
```

sudorandom:
Here are the hardware equivalents:
```
slcli --format=json call-api Account getHardware --mask="id,processorPhysicalCoreAmount,memoryCapacity,hostname,billingItem[id,orderItem[id,order[id,userRecord[id,username]]]]"
```
and
```
slcli call-api Account getHardware --mask="id,processorPhysicalCoreAmount,memoryCapacity,hostname,billingItem[id,orderItem[id,order[id,userRecord[id,username]]]]"
```

jravetch:
[...]

PR #614, opened 2015-08-18
server and vs lists now only mask in the visible fields

sudorandom:
This should resolve issues brought up in #547.

This also adds a 'tags' column


sudorandom:
Example showing all virtual servers and which user created them (requires order viewing permissions).

```
$ slcli vs list --columns=id,hostname,billingItem.orderItem.order.userRecord.username
```

underscorephil:
+1

Issue #620, opened 2015-08-27
It is not possible to upgrade ram


raulcabero:
When I try upgrading the ram for a vsi I got the following error: "Price does not have an id."

I ran the follwing comand:

$ slcli vs upgrade --memory 4 9637977
This action will incur charges on your account. Continue? [y/N]: y
SoftLayerAPIError(SoftLayer_Exception_Public): Price does not have an id.

It seems that the memory value is turned into megabytes, but that value does not match with the valid capacities listed by the getItems method and the result is an empty price id in the upgrade request

athak:
Even if the RAM value is specified in directly in megabytes, still get the same error.

underscorephil:
@raulcabero @athak: could you take a peek and verify that this is working properly for you?

raulcabero:
I am still getting the same error with the version 4.1.1 of the client running over a windows 7.

adamherzog:
@raulcabero This was only just fixed, and isn't included in the 4.1.1 release. It'll be included in the next release (4.1.2, I guess). In the meantime, you could install from the github repo and try it out.

(Instructions for installing from source here: https://softlayer-python.readthedocs.org/en/latest/install.html#from-source)

raulcabero:
I installed the client from the master branch now I was able to upgrade the ram in my VM successfully specifying the ram in gigabytes and megabytes .

Regards

PR #623, opened 2015-09-17
VS upgrade should use gb not mb for memory.

adamherzog:
This aims to address issue #620, where requests to upgrade the memory of a virtual server fails.

It seems that the CLI ensures that the memory is defined in megabytes, but the manager expects memory to be defined in gigabytes.


sudorandom:
Thanks!

PR #626, opened 2015-10-02
Covert to using pytest for the test runner

sudorandom:


briancline:
What are the advantages?

sudorandom:
@briancline  Mostly, better output when assertions fail.

sudorandom:
I also moved the tests out from under the SoftLayer module... If you hit any issues updating to this, run this to clear out __pycache__ and pyc files:

```bash
find . | grep -E "(__pycache__|\.pyc$)" | xargs rm -rf
```

sudorandom:
@underscorephil This is g2g. The coverage drop is expected... The coverage that it shows now is more accurate.

PR #640, opened 2015-11-17
Improve rest support, add total items to list results

### LINKED ISSUE #638: Total Items
When making API calls which has a listing of items, the API will return the total number of items in a header named 'SoftLayer-Total-Items'. This should be made available in the return value.

I propose wrapping API results that are lists with a class that looks like this:

```python
class SoftLayerListResult(list):
    def __init__(self, items, total_count):
        self.total_count = total_count
        super(SoftLayerListResult, self).__init__(items)
```

This should maintain backwards compatibility while exposing the total count of items to consumers.

### LINKED ISSUE #639: Resurrect masks/params/filters for REST transport
Work was done in a branch to add more of the normal features to the REST transport. This was abandoned because the use-case that prompted that work was satisfied in another way.

This issue is to track the work that adds masks, filters, limit, offsets, complex parameters to the SLAPI REST transport.

sudorandom:
* Resolves #638 - Lists now return from the client as SoftLayerListResult objects, which should behave like lists except for the fact that there is now a `total_count` attribute available which can tell you how many total items exist which satisfy your list query.
* Resolves #639 - Adds masks, filters, limit, offset, and complex parameters to the REST transport
* Fixes minor documentation issues
* Fixes issue when using the latest version of tox

### DIFF
--- a/SoftLayer/API.py
+++ b/SoftLayer/API.py
@@ -95,11 +95,21 @@ def create_client_from_env(username=None,
 
     # If we have enough information to make an auth driver, let's do it
     if auth is None and settings.get('username') and settings.get('api_key'):
+        # NOTE(kmcdonald): some transports mask other transports, so this is
+        # a way to find the 'real' one
+        real_transport = getattr(transport, 'transport', transport)
 
-        auth = slauth.BasicAuthentication(
-            settings.get('username'),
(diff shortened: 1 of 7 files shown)


briancline:
LGTM, thanks for adding this.

PR #654, opened 2015-12-02
CLI Usability Tweeks

sudorandom:
* **Removes global `--debug` option in favor of `--verbose` | `-v`.** -v can be specified multiple times to increase verbosity. For example, `-vvv` will the same as `--debug=3`
* Removes global `--timings` option
* Renames global `--fixtures` option to `--demo`
* Adds more diagnostics whenever the verbose (-v) flag is set including the info that was displayed with `--timings`
* Converts to using lowercase name/value for tables instead of upper-case since lower-case is generally preferred for this tool at the moment
* Shows default values for many CLI options

### DIFF
--- a/SoftLayer/CLI/cdn/detail.py
+++ b/SoftLayer/CLI/cdn/detail.py
@@ -17,9 +17,9 @@ def cli(env, account_id):
     manager = SoftLayer.CDNManager(env.client)
     account = manager.get_account(account_id)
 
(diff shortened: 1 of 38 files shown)


sudorandom:
The new pylint errors will be solved by #655

sudorandom:
This change is good to review.

underscorephil:
+1

PR #659, opened 2015-12-04
Adjusts how filters work with call-api. Adds --output-python option

sudorandom:
I'm not 100% on the change as is. There are a couple things I want to solve here:

 * Inserting JSON through a CLI command feels really bad to me. We're very close to just making REST calls with curl at this point. Therefore, I changed the way filters are defined. Instead of accepting JSON, you can specify filters with the format `-f 'property=value'`. You can also filter into relational properties with a dot-delimited format. E.G. `-f nested.property=value`. `-f` can be specified multiple times.
 * Since this is an experimentation tool, it should provide a way to convert these commands into real code. There seems to be a need for several features which go beyond this goal: complex arguments, filters, etc. For these cases we can do what we can but at a certain point code will need to be written either in another slcli command or with user-created scripts. To aid in this, the `--output-python` option was added to `call-api`. This outputs python code that shows an example of using softlayer-python to make the API call that `call-api` would have made. For example:

```
> slcli call-api Account getVirtualGuests \
                                         -f 'virtualGuests.datacenter.name=dal05' \
                                         -f 'virtualGuests.maxCpu=4' \
                                         --mask=id,hostname,datacenter.name,maxCpu test \
                                         --limit=10 \
                                         --output-python
import SoftLayer

client = SoftLayer.create_client_from_env()
result = client.call(u'Account',
                     u'getVirtualGuests',
                     u'test',
                     filter={u'virtualGuests': {u'datacenter': {u'name': {'operation': u'_= dal05'}}, u'maxCpu': {'operation': 4}}},
                     limit=1,
                     mask=u'id,hostname,datacenter.name,maxCpu')
```

### DIFF
--- a/SoftLayer/CLI/call_api.py
+++ b/SoftLayer/CLI/call_api.py
@@ -51,7 +51,8 @@ def _build_python_example(args, kwargs):
 @click.option('--id', '_id', help="Init parameter")
 @helpers.multi_option('--filter', '-f', '_filters',
                       help="Object filters. This should be of the form: "
-                      "'property=value' or 'nested.property=value'")
+                      "'property=value' or 'nested.property=value'. Complex "
+                      "filters like betweenDate are not currently supported.")
 @click.option('--mask', help="String-based object mask")
 @click.option('--limit', type=click.INT, help="Result limit")
 @click.option('--offset', type=click.INT, help="Result offset")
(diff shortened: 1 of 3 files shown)


allmightyspiff:
The -f option makes a lot of sense to me, and seems much more manageable than doing raw json.

My only complaint is that there doesn't seem to be any way to manage date filters in the command line. That might require a different switch though, due to how different filtering based off dates is from everything else.

if you decide not to include date filters, might be a good idea to mention that in the help section somewhere.

sudorandom:
I added a small note in the help text for `-f | --filter`

Issue #192, opened 2013-09-17
Enhance CCI DNS sync to update AAAA records


briancline:
Easy low-hanging fruit for the next milestone...

For folks with IPv6 addresses on their CCIs, having `sl cci dns sync` perform a sync of the AAAA record would be a useful addition as well.

Would be as simple as adding another `--AAAA` option to the existing command.

This would not include the full /64 block of addresses that is provided to them if they select at order time, only the *single* IPv6 address we provide to them if they chose it.

underscorephil:
@sudorandom @briancline: you can get the v6 address for a server with frontendNetworkComponents.primaryVersion6IpAddressRecord.

EX: https://gist.github.com/underscorephil/75709f051f4356f111e6

allmightyspiff:
https://github.com/softlayer/softlayer-python/pull/661

sudorandom:
Closed via #661

PR #661, opened 2015-12-08
issues192 added ability to sync aaaa records

allmightyspiff:
screen shot 2015-12-07 at 6 40 45 pm

### DIFF
--- a/SoftLayer/CLI/virt/dns.py
+++ b/SoftLayer/CLI/virt/dns.py
@@ -139,7 +139,7 @@ def cli(env, identifier, a_record, aaaa_record, ptr, ttl):
         raise exceptions.CLIAbort("Aborting DNS sync")
 
     both = False
-    if not ptr and not a_record:
+    if not ptr and not a_record and not aaaa_record:
         both = True
 
     if both or a_record:
--- a/SoftLayer/CLI/virt/dns.py
+++ b/SoftLayer/CLI/virt/dns.py
@@ -18,6 +18,9 @@ the records updated.""")
 @click.option('--a-record', '-a',
               is_flag=True,
               help="Sync the A record for the host")
+@click.option('--aaaa-record',
+              is_flag=True,
+              help="Sync the AAAA record for the host")
 @click.option('--ptr', is_flag=True, help="Sync the PTR record for the host")
 @click.option('--ttl',
               default=7200,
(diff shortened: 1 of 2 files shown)


SoftLayer/CLI/virt/dns.py
```diff
@@ -58,6 +72,38 @@ def sync_a_record():
...
+
+        if not records:
+            # don't have a record, lets add one to the base zone
+            dns.create_record(zone['id'],
+                              instance['hostname'],
+                              'aaaa',
+                              ip_address,
+                              ttl=ttl)
+        else:
+            recs = [x for x in records if x['type'].lower() == 'aaaa']
```
sudorandom: Do we not trust the record_type='aaaa' filter done when getting the dns records?

SoftLayer/CLI/virt/dns.py
```diff
@@ -58,6 +72,38 @@ def sync_a_record():
...
+            if len(recs) != 1:
+                raise exceptions.CLIAbort("Aborting A record sync, found "
```
sudorandom: s/A/AAAA

SoftLayer/CLI/virt/dns.py
allmightyspiff: I'm not sure, I just copied the logic from the create a record method. I'll see if using a filter would make the code look cleaner in the morning.

sudorandom: I was asking because you are already using a filter with the record_type argument and the A-record version doesn't. It appears like adding the filter made this line of code un-needed.

allmightyspiff:
@sudorandom fixed up those sections. Removed if statement from both, and both now have the appropriate filter.

This addresses https://github.com/softlayer/softlayer-python/issues/192

tests/CLI/modules/vs_tests.py
```diff
@@ -132,3 +131,35 @@ def test_create(self, confirm_mock):
...
+            {'id': 1, 
+             'ttl': 7200, 
+             'data': '172.16.240.2', 
+             'host': 'test', 
+             'type': 'a'}
+        ]
+        result = self.run_command(['vs', 'dns-sync', '100'])
+        self.assertEqual(result.exit_code, 0)
+        result = self.run_command(['vs', 'dns-sync', '--aaaa-record', '100'])
+        self.assertEqual(result.exit_code, 2)
```
sudorandom: You're going to want to assert that the proper API calls are made. Also, exit_code of 2 means something errorred. Typically you can print result.output or result.exception to see what happened with the command.

allmightyspiff: I've added checks for the proper API calls. And I do realize that exit_code 2 means there was an exception. I've added checks to make sure its CLIAbort exception instead of something else though.

I have also added some more checks to get the code coverage for this function to 100%.

allmightyspiff:
@sudorandom let me know if you'd like me to make any other updates on this.

sudorandom:
LGTM

PR #664, opened 2015-12-09
Update copyright year

sudorandom:
* Adds SoftLayerListResult to docs
 * Small formatting fix for the external links on the main page. Doing those links as a doctree was causing them to appear on the sidebar, which makes it really messy. This change converts it to just a list of links.


sudorandom:
I know the copyright year will have to update within a month, but this really should have been done a long time ago...

Issue #550, opened 2015-05-13
load balancer service-add command does not work through cli


fyatzeck:
The load balancer service-add command fails with error below while attempting to add service to the existing service group using port 15672.

Here is my command syntax:
slcli loadbal service-add local:79187 --port=15672 --weight=1 --healthcheck-type=4 --ip=174.37.219.102

I have also tried with ip address argument name spelled out as "--ip-address" resulting same error.  Is my syntax incorrect or is this a bug?

An unexpected error has occured:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/SoftLayer-4.0.0-py2.7.egg/SoftLayer/CLI/core.py", line 180, in main
    cli.main()
  File "/usr/local/lib/python2.7/site-packages/click-4.0-py2.7.egg/click/core.py", line 644, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/site-packages/click-4.0-py2.7.egg/click/core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/site-packages/click-4.0-py2.7.egg/click/core.py", line 991, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/site-packages/click-4.0-py2.7.egg/click/core.py", line 837, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/site-packages/click-4.0-py2.7.egg/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click-4.0-py2.7.egg/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click-4.0-py2.7.egg/click/core.py", line 464, in invoke
    return callback(*args, **kwargs)
TypeError: cli() got an unexpected keyword argument 'ip'

Feel free to report this error as it is likely a bug:
    https://github.com/softlayer/softlayer-python/issues

MiguelHigorre:
Hello,

It looks that there is an extra "-" character at line 27 in service_add.py module
@click.option('--ip-address', '--ip',
Changing that line to 
@click.option('--ip-address', '-ip',
It works for me, however the format for cli command sohuld be the following 
[...]

PR #668, opened 2015-12-18
Fixed Load Balancer Service Edit for CLI to support --disabled

toyray:
Removed `--ip` option (similar to fix for #550) and checked the options correctly so that `--disabled` is accepted as a valid option.

### DIFF
--- a/SoftLayer/CLI/loadbal/service_edit.py
+++ b/SoftLayer/CLI/loadbal/service_edit.py
@@ -20,7 +20,7 @@ from SoftLayer.CLI import loadbal
               type=click.INT,
               help="Change the weight of the service")
 @click.option('--healthcheck-type', help="Change the health check type")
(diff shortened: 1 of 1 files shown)


SoftLayer/CLI/loadbal/service_edit.py
```diff
@@ -20,7 +20,7 @@
               type=click.INT,
               help="Change the weight of the service")
 @click.option('--healthcheck-type', help="Change the health check type")
-@click.option('--ip-address', '--ip', help="Change the IP of the service")
+@click.option('--ip-address', help="Change the IP of the service")
```
underscorephil: can we change this to "IP address" to match the verbiage in #562 ?

sudorandom:
+1

toyray:
@underscorephil, updated the description for the `--ip-address` option as requested :grin:

Issue #681, opened 2016-01-26
VS upgrade validation


underscorephil:
slcli vs upgrade needs validation to prevent the attempt of an empty upgrade:
```
(venv)❯ ./venv/bin/slcli vs upgrade 14354931
This action will incur charges on your account. Continue? [y/N]: y
VS Upgrade Failed
```

At least one upgrade parameter should be required for the command to proceed.

PR #684, opened 2016-02-03
Check at least one upgrade option is provided.

camporter:
Ensure that private option requires the cpu option.
Filter out location based prices when upgrading.
Add additional tests around guest upgrading.

Addresses #681


sudorandom:
+1

PR #696, opened 2016-03-14
Fixed helper text

greyhoundforty:
Currently the helper text for the server reload function has the descriptions swapped:

```
❯ slcli server reload --help
Usage: slcli server reload [OPTIONS] IDENTIFIER

  Reload operating system on a server.

Options:
  -i, --postinstall TEXT  SSH keys to add to the root user
  -k, --key TEXT          Post-install script to download
                          (Only HTTPS
                          executes, HTTP leaves file in /root) (multiple
                          occurrence permitted)
  -h, --help              Show this message and exit
```.

### DIFF
--- a/SoftLayer/CLI/hardware/reload.py
+++ b/SoftLayer/CLI/hardware/reload.py
@@ -12,9 +12,9 @@ from SoftLayer.CLI import helpers
 
 @click.command()
 @click.argument('identifier')
-@click.option('--postinstall', '-i',
-              help=("Post-install script to download "
-                    "(Only HTTPS executes, HTTP leaves file in /root"))
+@helpers.multi_option('--postinstall', '-i',
+                      help=("Post-install script to download "
+                            "(Only HTTPS executes, HTTP leaves file in /root"))
 @helpers.multi_option('--key', '-k', help="SSH keys to add to the root user")
 @environment.pass_env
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: SoftLayer/CLI/hardware/reload.py [hunk 1]
```diff
@@ -12,9 +12,9 @@
 
 @click.command()
 @click.argument('identifier')
-@click.option('--postinstall', '-i',
-              help=("Post-install script to download "
-                    "(Only HTTPS executes, HTTP leaves file in /root"))
+@helpers.multi_option('--postinstall', '-i',
```
sudorandom: Wait, I don't think you can give multiple post install scripts when reloading hardware. SoftLayer_Container_Hardware_Server_Configuration.customProvisionScriptUri takes in a string, not an array.

ref:
 - http://sldn.softlayer.com/reference/services/SoftLayer_Hardware_Server/reloadOperatingSystem
 - http://sldn.softlayer.com/reference/datatypes/SoftLayer_Container_Hardware_Server_Configuration

greyhoundforty:
Well the helper text needs to be fixed and @allmightyspiff mentioned that the `helpers.multi_option` needed to be there for the helper text to be correct.

sudorandom:
@greyhoundforty the text is already fixed in master.

Issue #701, opened 2016-03-20
SoftLayer.managers.cdn.purge_content throws a purgeContent does not exist for service error


alainrobert:
### Expected Behavior

self._cdn.purge_content(self._cdn_account, url )  
returns True or message
 
### Actual Behavior

self._cdn.purge_content(self._cdn_account, url )  
throws Function ("purgeContent") is not a valid method for this service

### Environment Information
Operating System:
softlayer-python version (`slcli --version`):  4.1.1 and 5.0.0

In the code replace  self.account.purgeContent with self.account.purgeCache seems to solve the problem.

sudorandom:
Resolved in v5.0.1

PR #702, opened 2016-03-20
Updates CDN manager to use new purge method

sudorandom:
Resolves issue #701


briancline:
Looks reasonable to me.

Issue #667, opened 2015-12-14
Ticket, attach hardware/virtual machine


underscorephil:
Add support to add/list/remove ticket to device relationships

# CLI
```
slcli ticket create --title='plz halp' --subject-id=$SUBJECTID --body='me' --device=1234
```
We may have to add in switches to differentiate hardware/vs
```
slcli ticket create --title='plz halp' --subject-id=$SUBJECTID --body='me' --hardware=1234
slcli ticket create --title='plz halp' --subject-id=$SUBJECTID --body='me' --virtual=1234
```
# Referenced methods
## Virtual
http://sldn.softlayer.com/reference/services/SoftLayer_Ticket/addAttachedVirtualGuest
http://sldn.softlayer.com/reference/services/SoftLayer_Ticket/removeAttachedVirtualGuest
http://sldn.softlayer.com/reference/services/SoftLayer_Ticket/getAttachedVirtualGuests
## Hardware
http://sldn.softlayer.com/reference/services/SoftLayer_Ticket/addAttachedHardware
http://sldn.softlayer.com/reference/services/SoftLayer_Ticket/removeAttachedHardware
http://sldn.softlayer.com/reference/services/SoftLayer_Ticket/getAttachedHardware

camporter:
This issue should be addressed in master now :shipit:

PR #715, opened 2016-04-04
Add ability to attach/detach devices to tickets.

camporter:
Add attach arguments for hardware and virtual to ticket create CLI.
Add new ticket attach and detach CLI commands.
Add methods to ticket manager to attach and detach devices.

Addresses #667


sudorandom:
LGTM

Issue #727, opened 2016-05-25
Add "in" filter for call-api command


sudorandom:
### Expected Behavior
When making this call:
```
slcli call-api Account getVirtualGuests -f 'virtualGuests.domain=example.com,example2.com'
```
This should return a list of virtual guests that have either example.com or example2.com as a domain

### Actual Behavior
It returns a list of virtual guests that has "example.com,example2.com" as a domain.

Obviously, if you want to filter for something with commas in it, this will be an issue. However, I don't think that case happens often enough to worry about it.

sudorandom:
An alternative approach is this:

```slcli call-api Account getVirtualGuests -f 'virtualGuests.domain IN example.com,example2.com'```

No matter what, we either have to assign an escape character or not worry about edge-cases because the string " IN " can also appear in names. I'm worried about this becoming too much like its own language. That is why I still think the original proposal is still acceptable. Commas in names of things you want to filter on seems exceedingly rare and this tool is for experimentation of the API... not for actual production use.

cc: @camporter.

camporter:
Yeah, I suppose if we don't enforce spaces around the operators, finding the 'IN' operator isn't trivial to do and could occur in other parts of the filter unintentionally. The grammar for the CLI should be kept simple to use and implement.

sudorandom:
Closed due to #728.

PR #728, opened 2016-05-25
Adds support for the 'in' filter to slcli call-api

sudorandom:


sudorandom:
Resolves #727

thekuffs:
lgtm :shipit:

camporter:
A more backwards-compatible way of changing this would be to add a new keyword. The docs for the _build_filters method says "This only supports the equals keyword at the moment", which seems to indicate that additional keywords would be added in the future.

There could be cases, such as various names on objects (images, ssh keys, etc), that may have commas in them.

Issue #731, opened 2016-06-08
Feature Request: Add functionality to authorize hosts with slcli block


lukecoll:
### Expected Behavior
An extension of the item https://github.com/softlayer/softlayer-python/issues/729
We would like there to be a command which allows us to authorize hosts to storage using the CLI.

### Actual Behavior
Unable to authorize hosts with the slcli block command(s). Task must be completed via Web UI.

### Environment Information
Operating System:
softlayer-python version (`slcli --version`):
slcli (SoftLayer Command-line), version 5.1.0

Please let me know if we opened this incorrectly.

lukecoll:
Also if this is available via other API implementations please let me know so we can implement a stop-gap on our side.

vmtyler:
The code exists here: https://github.com/iodine53/softlayer-python/tree/block-storage-host-auth
I'm working on test coverage and will submit it as a PR

sudorandom:
Closed with #742

PR #742, opened 2016-07-23
Block storage host auth

vmtyler:
Pulled forward the 4.1.0 branch from @iodine53 and cleaned up a few things. Adds ability to give and revoke access to block volumes, per https://github.com/softlayer/softlayer-python/issues/731

### DIFF
--- a/SoftLayer/CLI/block/access/revoke.py
+++ b/SoftLayer/CLI/block/access/revoke.py
@@ -1,0 +1,38 @@
+"""Create a block storage snapshot."""
+# :license: MIT, see LICENSE for more details.
+
+import click
+import SoftLayer
+from SoftLayer.CLI import environment
+
+
+@click.command()
+@click.argument('volume_id')
+@click.option('--hardware_id', '-h', multiple=True,
+              help='The id of one SoftLayer_Hardware'
+              ' to revoke authorization')
+@click.option('--virtual_id', '-v', multiple=True,
+              help='The id of one SoftLayer_Virtual_Guest'
+              ' to revoke authorization')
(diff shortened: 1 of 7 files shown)


SoftLayer/managers/block.py
```diff
@@ -135,6 +135,80 @@ def get_block_volume_snapshot_list(self, volume_id, **kwargs):
...
+                })
+
+        if ip_address_ids is not None:
+            for ip_address_id in ip_address_ids:
+                host_templates.append({
+                    'objectType': 'SoftLayer_Network_Subnet_IpAddress',
+                    'id': ip_address_id
+                })
+
+        return self.client['Network_Storage'].\
```
sudorandom: Small nit: This is a lot more natural like this:

```python
        return self.client.call('Network_Storage', 'allowAccessFromHostList',
                                host_templates, id=volume_id)
```

SoftLayer/CLI/block/access/revoke.py
```diff
@@ -0,0 +1,38 @@
...
+# :license: MIT, see LICENSE for more details.
+
+import click
+import SoftLayer
+from SoftLayer.CLI import environment
+
+
+@click.command()
+@click.argument('volume_id')
+@click.option('--hardware_id', '-h', multiple=True,
```
sudorandom: Use dashes in favor of underscores for CLI options. `--hardware-id` over `--hardware_id`.

vmtyler:
Ok got those fixes in. Fixed the dashes in the authorize CLI as well a the revoke CLI.

PR #748, opened 2016-08-04
Add File commands and new Block commands

dpickle2:
Cherrypick manual Block snapshot creation from iodine53's branch
Additional code developed collaboratively by:
    ccorales95  Cristina Corales <ccorales95@gmail.com>
    kyubifire   Nilo Lisboa <nilo_bl@hotmail.com>
    VuKevin     Kevin Vu <VuKevin@icloud.com>
    dpickle2    David Pickle <dpickle29@gmail.com>

* Add support for manual snapshots [Block & File]
  (cherrypick from iodine53's branch for Block, then
   based similar function for File on this code)

* Add support for deleting snapshots [File]

* Add support for listing snapshots [File]

* Add snapshot schedule enable/disable [Block & File]

* Add snapshot space ordering [Block & File]

* Add restoring a volume to snapshots [Block & File]

* Add cancelling snapshot space [Block & File]

* Add storage host authorization [File]

* Add listing hosts authorized to storage volumes [File]

* Add ordering File volumes [File]

* Add cancelling File volumes [File]

* Add displaying File volume details and volume lists [File]

* Add additional output for commands:
      Add used space (bytes) to detail view  [File]
      Add LUN ID column for block volume-list [Block]
      Add "success output" for several commands [Block & File]
      Add transaction info to detail/list views [Block & File]

* Fix block snapshot-list command error
  (The "snapshot-list" block command produced an error when the
  list of snapshots for a volume was not empty. Updating the object
  mask which is sent to the SLAPI fixed the problem.)

* Create "storage_utils.py" files for duplicate code
  (To prevent duplicate code between File and Block commands,
  code was refactored into SoftLayer/managers/storage_utils.py
  and SoftLayer/CLI/storage_utils.py.)


SoftLayer/CLI/block/cancel.py
```diff
@@ -25,4 +25,15 @@ def cli(env, volume_id, reason, immediate):
...
+
+    if cancelled:
+        if immediate:
+            click.echo('Block volume with id %s has been marked'
+                       ' for immediate cancellation' % volume_id)
+        else:
+            click.echo('Block volume with id %s has been marked'
+                       ' for cancellation' % volume_id)
+    else:
+        click.echo('Unable to cancle block volume %s' % volume_id)
```
camporter: cancel*

SoftLayer/CLI/block/snapshot/disable.py
```diff
@@ -0,0 +1,29 @@
...
+              help='Snapshot schedule [HOURLY|DAILY|WEEKLY]',
+              required=True)
+@environment.pass_env
+def cli(env, volume_id, schedule_type):
+    """Disables snapshots on the specified schedule for a given volume"""
+
+    if (schedule_type != 'HOURLY' and schedule_type != 'DAILY'
+            and schedule_type != 'WEEKLY'):
+        raise exceptions.CLIAbort(
+            '--schedule_type must be HOURLY, DAILY, or WEEKLY')
```
camporter: --schedule-type

SoftLayer/CLI/file/list.py
```diff
@@ -0,0 +1,69 @@
...
+
+COLUMNS = [
+    column_helper.Column('id', ('id',), mask="id"),
+    column_helper.Column('username', ('username',), mask="username"),
+    column_helper.Column('datacenter',
+                         ('serviceResource', 'datacenter', 'name'),
+                         mask="serviceResource.datacenter.name"),
+    column_helper.Column(
+        'storage_type',
+        lambda b: b['storageType']['keyName'].split('_').pop(0),
```
underscorephil: Is there any situation which this value would not be returned by the API properly?

SoftLayer/CLI/file/snapshot/enable.py
```diff
@@ -0,0 +1,58 @@
...
+@environment.pass_env
+def cli(env, volume_id, schedule_type, retention_count,
+        minute, hour, day_of_week):
+    """Enables snapshots for a given volume on the specified schedule"""
+    file_manager = SoftLayer.FileStorageManager(env.client)
+
+    valid_days = {'SUNDAY', 'MONDAY', 'TUESDAY', 'WEDNESDAY', 'THURSDAY',
+                  'FRIDAY', 'SATURDAY'}
+
+    if (schedule_type != 'HOURLY' and schedule_type != 'DAILY'
```
underscorephil: Let's move this to a list with 'if in' as we did with day of the week to keep it consistent

dpickle2:
Alright, I updated the checks for the schedule type to be more internally consistent in both the File and Block versions of the snapshot-enable command.

Regarding the lambda function in the volume-list commands:
So long as the API always returns a SoftLayer_Network_Storage object with "storageType" and "storageType.keyName" present, I would think it would be okay, but we were not certain if this would always be true. So, I modified the lambda function to filter out several different invalid cases. Please let me know if this approach is okay. (For reference, the original version of the lambda function was taken from the previously existing "block volume-list" command.)

PR #752, opened 2016-08-17
Add File & Block replication commands and output

dpickle2:
This pull request includes several new commands and updates
to CLI output for supporting File and Block volume replication.

Code developed collaboratively by:
    ccorales95  Cristina Corales <ccorales95@gmail.com>
    kyubifire   Nilo Lisboa <nilo_bl@hotmail.com>
    VuKevin     Kevin Vu <VuKevin@icloud.com>
    dpickle2    David Pickle <dpickle29@gmail.com>

* Add support for ordering replicant volumes [Block & File]

* Add support for replicant failover & failback [Block & File]

* Add additional output for commands:
    Add replicant info to volume-detail output [Block & File]
    Show replicant volumes in volume-list output [Block & File]


SoftLayer/CLI/block/replication/failback.py
```diff
@@ -0,0 +1,23 @@
...
+def cli(env, volume_id, replicant_id):
+    """Failback a block volume from the given replicant volume."""
+    block_storage_manager = SoftLayer.BlockStorageManager(env.client)
+
+    success = block_storage_manager.failback_from_replicant(
+        volume_id,
+        replicant_id
+    )
+
+    if success:
```
sudorandom: If success is `False`, shouldn't we return some kind of error message to the user? I'm not sure what conditions `False` would be returned over returning an exception.

SoftLayer/CLI/file/replication/order.py
```diff
@@ -0,0 +1,53 @@
+"""Order a file storage replica volume."""
+# :license: MIT, see LICENSE for more details.
+
+import click
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import exceptions
+
+
+CONTEXT_SETTINGS = dict(token_normalize_func=lambda x: x.upper())
```
sudorandom: The dict syntax is preferred here. It's a bit shorter and slightly faster:
```python
CONTEXT_SETTINGS ={'token_normalize_func': lambda x: x.upper()}
```

SoftLayer/managers/block.py
```diff
@@ -184,6 +190,44 @@ def deauthorize_host_to_volume(self, volume_id,
...
+        """
+
+        block_mask = 'billingItem[activeChildren],storageTierLevel,'\
+                     'osType,snapshotCapacityGb,schedules,'\
+                     'hourlySchedule,dailySchedule,weeklySchedule'
+        block_volume = self.get_block_volume_details(volume_id,
+                                                     mask=block_mask)
+
+        if os_type is None:
+            if 'osType' in block_volume\
```
sudorandom: There is a utility function in SoftLayer.utils for looking up things in a nested dict much easier. For example:
```
        if isinstance(utils.lookup(block_volume, 'osType', 'keyName'), str):
```

Also, all this logic is to make os_type optional. Typically that is done by making os_type a keyword argument with a default value. In this case, the appropriate default is probably `None`.

dpickle2:
Thank you for your feedback! I have addressed the items you mentioned above. Please let me know if I have missed anything else.

[The failover/failback commands resulted in exceptions raised by the SL API for us, but I have added else clauses to each of these commands to address the possibility of False being returned by the API.]

[Thank you for pointing out the use of default values for arguments. I have updated the order_replicant_volume() functions in both File and Block managers to make 'tier' and 'os_type' optional with default values.]

PR #754, opened 2016-08-17
Fix issue 753 to add "created_by" column to block / file

mjfork:


SoftLayer/CLI/file/list.py
```diff
@@ -27,6 +27,9 @@
                          mask="serviceResourceBackendIpAddress"),
     column_helper.Column('active_transactions', ('activeTransactionCount',),
                          mask="activeTransactionCount"),
+    column_helper.Column(
```
sudorandom: You probably want it to look like this:

```python

    column_helper.Column(
        'created_by',
        ('billingItem', 'orderItem', 'order', 'userRecord', 'username')),
```

Issue #755, opened 2016-08-17
"Unable to add a Graphics Processing Unit price" on new bare metal hardware create


kramimus:
### Expected Behavior
```
slcli hw create -H test -D <domain> -s S1270_8GB_2X1TBSATA_NORAID -o UBUNTU_16_64 -d hou02 --port-speed 100 --billing hourly -k <key>
```
Should launch a new hourly bare metal server

### Actual Behavior
```
$ slcli hw create -H test -D <domain> -s S1270_8GB_2X1TBSATA_NORAID -o UBUNTU_16_64 -d hou02 --port-speed 100 --billing hourly -k <key>
This action will incur charges on your account. Continue? [y/N]: y
SoftLayerAPIError(SoftLayer_Exception_Public): Unable to add a Graphics Processing Unit price (178117) because it is not valid for the package (200).
```

### Environment Information
Operating System: Ubuntu 16.04
softlayer-python version (`slcli --version`): 5.1.0

I was able to create a server without issue yesterday with the exact same command. I do not want a GPU on the hardware I am trying to order.

kramimus:
I dumped the price list being sent to the API and the above price ID is not listed, so I assume this is probably an API problem, not something on the client.

sudorandom:
I'm re-opening this. I believe a change was made on the API side that is possibly forcing a change here. I saw that there is more discussion happening about this issue here: https://stackoverflow.com/questions/39003084/gpu-related-api-errors-for-softlayer-hourly-bare-metal-server-creation.

kramimus:
I updated the SO issue, but I am no longer getting this API error and my create requests are going through.

sudorandom:
Thanks for the update. I'm closing this issue now.

PR #757, opened 2016-08-17
Issue 755

sudorandom:
Resolves issue #755. I have no idea why there's an item named "None" with the category of "other" and I have no idea why this fixes the gpu price issue... however... it does seem fix it.


sudorandom:
Closing this. It seems that this price is now added as a default.

PR #763, opened 2016-08-23
CLI: Makes the price shown more accurate for virtual and hardware detail

### LINKED ISSUE #749: slcli vs detail servername --price is returning the wrong price
### Expected Behavior
I would expect 
slcli vs detail servername --price
to return the full price of the server.
### Actual Behavior
From https://control.softlayer.com/support/tickets/31456403
The 'price rate' that you retrieved using the API is actually the line item pricing for just the processor of the server, while the price that you see in the next invoice is the total cost for the server. At this point in time, we are not sure how you are going about using the API to retrieve the billing information, but the server on the whole is considered the "parent item", which you may have missed out on earlier. 

Kindly refer to this page and make the adjustments, this time with the server as "parent item" in your consideration.
<http://sldn.SoftLayer.com/reference/services/SoftLayer_Billing_Item>

Best regards,
Ernest T.
Digital Account Sales
### Environment Information
Operating System: OSX
softlayer-python version (`slcli --version`): slcli (SoftLayer Command-line), version 4.1.1

sudorandom:
Incorporates the child prices into the price for virtual and hardware detail.
This impacts:
	`slcli virtual detail [ID] --price`
	`slcli hardware detail [ID] --price`

It also adjusts the default masks for detailing hardware and virtual instances.

Side note: I am on the edge of removing this feature now and coming back to it later with more validation. The biggest issue that I have testing this is that I don't have a normal account and it makes testing parts of the API that have prices harder.

Resolves #749

### DIFF
--- a/SoftLayer/CLI/hardware/detail.py
+++ b/SoftLayer/CLI/hardware/detail.py
@@ -81,10 +81,15 @@ def cli(env, identifier, passwords, price):
         table.add_row(['notes', result['notes']])
 
     if price:
-        table.add_row(['price rate',
-                       utils.lookup(result,
-                                    'billingItem',
-                                    'nextInvoiceTotalRecurringAmount')])
+        total_price = utils.lookup(result,
+                                   'billingItem',
+                                   'nextInvoiceTotalRecurringAmount') or 0
+        total_price += sum(p['nextInvoiceTotalRecurringAmount']
+                           for p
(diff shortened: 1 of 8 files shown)


sudorandom:
cc: @imacdonald-cb

imacdonald-cb:
Let me know if you want me to do some additional testing?

PR #769, opened 2016-08-30
Enable zone import to import wildcard entries

alexderz:
DNS CLI tests pass


sudorandom:
LGTM. Thanks for the PR!

PR #774, opened 2016-09-21
Added trunk option to hardware edit

nanjj:
to add network vlan trunk support in CLI

### DIFF
--- a/SoftLayer/CLI/hardware/edit.py
+++ b/SoftLayer/CLI/hardware/edit.py
@@ -28,9 +28,14 @@ from SoftLayer.CLI import helpers
               help="Private port speed.",
               default=None,
               type=click.Choice(['0', '10', '100', '1000', '10000']))
+@click.option('--trunk',
+              help="Add or remove vlan trunk, positive vlan id "
+              "means to add, negative means to remove",
+              default=None,
+              type=click.INT)
 @environment.pass_env
 def cli(env, identifier, domain, userfile, tag, hostname, userdata,
-        public_speed, private_speed):
+        public_speed, private_speed, trunk):
     """Edit hardware details."""
 
     if userdata and userfile:
@@ -62,3 +67,9 @@ def cli(env, identifier, domain, userfile, tag, hostname, userdata,
 
     if private_speed is not None:
         mgr.change_port_speed(hw_id, False, int(private_speed))
+
+    if trunk is not None:
+        if not mgr.trunk(hw_id, vlan_id=trunk):
+            raise exceptions.CLIAbort(
+                'Failed to %s vlan trunk %s' % (
+                    'add' if trunk > 0 else 'remove', abs(trunk)))
--- a/SoftLayer/managers/hardware.py
+++ b/SoftLayer/managers/hardware.py
@@ -50,16 +50,16 @@ class HardwareManager(utils.IdentifierMixin, object):
         else:
             self.ordering_manager = ordering_manager
 
-    def cancel_hardware(self, hardware_id, reason='unneeded', comment='',
+    def cancel_hardware(self, network_id, reason='unneeded', comment='',
                         immediate=False):
         """Cancels the specified dedicated server.
 
         Example::
 
             # Cancels hardware id 1234
-            result = mgr.cancel_hardware(hardware_id=1234)
+            result = mgr.cancel_hardware(network_id=1234)
 
-        :param int hardware_id: The ID of the hardware to be cancelled.
+        :param int network_id: The ID of the hardware to be cancelled.
         :param string reason: The reason code for the cancellation. This should
                               come from :func:`get_cancellation_reasons`.
         :param string comment: An optional comment to include with the
@@ -70,7 +70,7 @@ class HardwareManager(utils.IdentifierMixin, object):
         reasons = self.get_cancellation_reasons()
         cancel_reason = reasons.get(reason, reasons['unneeded'])
 
-        hw_billing = self.get_hardware(hardware_id,
+        hw_billing = self.get_hardware(network_id,
                                        mask='mask[id, billingItem.id]')
         if 'billingItem' not in hw_billing:
             raise SoftLayer.SoftLayerError(
@@ -169,7 +169,7 @@ class HardwareManager(utils.IdentifierMixin, object):
         kwargs['filter'] = _filter.to_dict()
         return self.account.getHardware(**kwargs)
 
-    def get_hardware(self, hardware_id, **kwargs):
+    def get_hardware(self, network_id, **kwargs):
         """Get details about a hardware device.
 
         :param integer id: the hardware ID
@@ -180,7 +180,7 @@ class HardwareManager(utils.IdentifierMixin, object):
 
             object_mask = "mask[id,networkVlans[vlanNumber]]"
             # Object masks are optional
-            result = mgr.get_hardware(hardware_id=1234,mask=object_mask)
+            result = mgr.get_hardware(network_id=1234,mask=object_mask)
         """
 
         if 'mask' not in kwargs:
@@ -224,12 +224,12 @@ class HardwareManager(utils.IdentifierMixin, object):
                 'remoteManagementAccounts[username,password]'
             )
 
-        return self.hardware.getObject(id=hardware_id, **kwargs)
+        return self.hardware.getObject(id=network_id, **kwargs)
 
-    def reload(self, hardware_id, post_uri=None, ssh_keys=None):
+    def reload(self, network_id, post_uri=None, ssh_keys=None):
         """Perform an OS reload of a server with its current configuration.
 
-        :param integer hardware_id: the instance ID to reload
+        :param integer network_id: the instance ID to reload
         :param string post_url: The URI of the post-install script to run
                                 after reload
         :param list ssh_keys: The SSH keys to add to the root user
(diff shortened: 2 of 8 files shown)


SoftLayer/managers/hardware.py
```diff
@@ -257,6 +257,28 @@ def rescue(self, hardware_id):
...
+
+            result = mgr.trunk(824671, 1234)
+        """
+        hw = self.get_hardware(hardware_id)
+        primary = 'primaryBackendIpAddress'
+        if public:
+            primary = 'primaryFrontendIpAddress'
+        cids = [c['id'] for c in hw['networkComponents']
+                if c.get('primaryIpAddress') == hw[primary]]
+        param = [{'vlanNumber': int(vlan_id)}]
```
sudorandom: The documentation expects an integer for `vlan_id`. Casting like this should not be needed.

SoftLayer/managers/hardware.py
```diff
@@ -257,6 +257,28 @@ def rescue(self, hardware_id):
         """
         return self.hardware.bootToRescueLayer(id=hardware_id)
 
+    def trunk(self, hardware_id, public, vlan_id):
+        """Add vlan trunk to a server.
+
+        :param integer hardware_id: the hardware ID to add vlan trunk
+        :param integer vlan_id: the vlan number
+
```
sudorandom: The `public` parameter is missing from the docblock

SoftLayer/managers/hardware.py
```diff
@@ -257,6 +257,28 @@ def rescue(self, hardware_id):
...
+        """
+        hw = self.get_hardware(hardware_id)
+        primary = 'primaryBackendIpAddress'
+        if public:
+            primary = 'primaryFrontendIpAddress'
+        cids = [c['id'] for c in hw['networkComponents']
+                if c.get('primaryIpAddress') == hw[primary]]
+        param = [{'vlanNumber': int(vlan_id)}]
+        if len(cids) == 0:
+            return
```
sudorandom: I feel like this is almost always an error. "You tried to trunk an interface that doesn't exist"

SoftLayer/managers/hardware.py
```diff
@@ -257,6 +257,28 @@ def rescue(self, hardware_id):
...
+
+        Example::
+
+            result = mgr.trunk(824671, 1234)
+        """
+        hw = self.get_hardware(hardware_id)
+        primary = 'primaryBackendIpAddress'
+        if public:
+            primary = 'primaryFrontendIpAddress'
+        cids = [c['id'] for c in hw['networkComponents']
```
sudorandom: You can avoid doing this awkward filtering by masking in primaryNetworkComponent or primaryBackendNetworkComponent instead of using get_hardware().

E.G.:
```
hw = self.client.call('Hardware', 'getObject', mask='id,primaryNetworkComponentprimaryBackendNetworkComponent')
```

SoftLayer/CLI/hardware/edit.py
```diff
@@ -62,3 +70,9 @@ def cli(env, identifier, domain, userfile, tag, hostname, userdata,
 
     if private_speed is not None:
         mgr.change_port_speed(hw_id, False, int(private_speed))
+
+    if private_trunk is not None:
+        mgr.trunk(hw_id, False, int(private_trunk))
```
sudorandom: I think this int() casting is not needing since the type the click library already validates for an int.

SoftLayer/managers/hardware.py
```diff
@@ -257,6 +257,28 @@ def rescue(self, hardware_id):
         """
         return self.hardware.bootToRescueLayer(id=hardware_id)
 
+    def trunk(self, hardware_id, public, vlan_id):
```
sudorandom: This method has no tests.

SoftLayer/CLI/hardware/edit.py
```diff
@@ -62,3 +70,9 @@ def cli(env, identifier, domain, userfile, tag, hostname, userdata,
...
+
+    if public_trunk is not None:
+        mgr.trunk(hw_id, True, int(public_trunk))
```
sudorandom: I think this int() casting is not needing since the type the click library already validates for an int.

SoftLayer/CLI/hardware/edit.py
```diff
@@ -62,3 +70,9 @@ def cli(env, identifier, domain, userfile, tag, hostname, userdata,
 
     if private_speed is not None:
         mgr.change_port_speed(hw_id, False, int(private_speed))
+
+    if private_trunk is not None:
+        mgr.trunk(hw_id, False, int(private_trunk))
+
+    if public_trunk is not None:
```
sudorandom: There's no test which hits this condition.

SoftLayer/CLI/hardware/edit.py
```diff
@@ -62,3 +70,9 @@ def cli(env, identifier, domain, userfile, tag, hostname, userdata,
 
     if private_speed is not None:
         mgr.change_port_speed(hw_id, False, int(private_speed))
+
+    if private_trunk is not None:
```
sudorandom: There's no test which hits this condition.

SoftLayer/managers/hardware.py
```diff
@@ -257,6 +257,28 @@ def rescue(self, hardware_id):
...
+        """Add vlan trunk to a server.
+
+        :param integer hardware_id: the hardware ID to add vlan trunk
+        :param integer vlan_id: the vlan number
+
+        Example::
+
+            result = mgr.trunk(824671, 1234)
```
sudorandom: This example doesn't match the current method signature.

SoftLayer/managers/hardware.py
sudorandom: If we're already requiring the vlan_id to trunk, can't we  look up if this is a public or private vlan and trunk to the correct network interface using that information?

SoftLayer/managers/hardware.py
nanjj: Added

SoftLayer/managers/hardware.py
nanjj: Fixed

SoftLayer/managers/hardware.py
nanjj: Fixed

SoftLayer/managers/hardware.py
nanjj: fixed

SoftLayer/managers/hardware.py
nanjj: fixed

SoftLayer/CLI/hardware/edit.py
nanjj: fixed

SoftLayer/managers/hardware.py
nanjj: Done.

nanjj: Done

nanjj:
So all comments have been fixed, please review again.

SoftLayer/managers/hardware.py
```diff
@@ -50,16 +50,16 @@ def __init__(self, client, ordering_manager=None):
         else:
             self.ordering_manager = ordering_manager
 
-    def cancel_hardware(self, hardware_id, reason='unneeded', comment='',
+    def cancel_hardware(self, network_id, reason='unneeded', comment='',
```
sudorandom: You seam to have renamed `hardware_id` to `network_id` in the entire hardware manager file... That doesn't seem to make sense. Find/replace error?

SoftLayer/managers/hardware.py
```diff
@@ -255,12 +255,41 @@ def rescue(self, hardware_id):
...
+
+        :param integer hardware_id: the hardware ID to add vlan trunk.
+        :param integer vlan_id: the vlan id.
+
+        Example::
+
+            # Add trunk
+            result = mgr.trunk(824671, 1341234)
+            # Remove trunk
+            result = mgr.trunk(824671, -1341234)
```
sudorandom: Although this is neat, it's not at all obvious that making the vlan id negative will remove it... How I would do this is make two methods: `add_trunk` and `remove_trunk` so it's obvious what kind of action is happening. In the CLI, you could have `--remove-trunk=1234` and `--add-trunk=4321` options.

SoftLayer/managers/hardware.py
```diff
@@ -255,12 +255,41 @@ def rescue(self, hardware_id):
...
+        action = 'addNetworkVlanTrunks'
+        if vlan_id < 0:
+            action = 'removeNetworkVlanTrunks'
+        vlan_id = abs(vlan_id)
+        vmgr = SoftLayer.NetworkManager(self.client)
+        vlan = vmgr.get_vlan(vlan_id)
+        name = "primaryBackendNetworkComponent"
+        if vlan['networkSpace'] == 'PUBLIC':
+            name = "primaryNetworkComponent"
+        mask = 'id,' + name + ',' + name
```
sudorandom: name is in the mask twice?

sudorandom:
@nanjj, before you start responding to the notes I just made we should have a discussion in a non-code issue about this feature and then implement it in one or more issues.

There are some high level issues that I just found out from a networking dev that will cause this to have issues. You have to know specifically which network interface you want to trunk since there are often times two network interfaces for the network. For example, a lot of setups have dual path for both public and private traffic so there would be 4 network interfaces in total for normal traffic. There's also a decision about putting this in the vlan part of the CLI or the hardware/virtual side.

So please, in another issue describe what use-cases you're trying to cover with this and what you want the CLI and manager signatures to look like.

cc: @benlake

PR #781, opened 2016-10-28
Changes to File volume list and detail commands to display NFS MounPoint for Volumes

skraghu:
This change would enable the `slcli file volume-list` and `slcli file volume-detail` commands to display the NFS Mount Point Address for each volume that it shows.


sudorandom:
LGTM

Issue #809, opened 2017-03-30
Assuming a single package with a bare metal type is problematic


kwienken:
The hardware manager's get_package assumes there will only be a single package with the BARE_METAL_CPU_FAST_PROVISION type. This becomes problematic if new packages are introduced with that package type. The hardware create and create-options will fail with "Ordering package not found".

PR #810, opened 2017-03-30
Add getting package by key for use in hardware manager

kwienken:
This addresses the issue I opened, #809.

### DIFF
--- a/SoftLayer/managers/hardware.py
+++ b/SoftLayer/managers/hardware.py
@@ -411,13 +411,13 @@ activePresets,
 regions[location[location[priceGroups]]]
 '''
 
-        package_type = 'BARE_METAL_CPU_FAST_PROVISION'
-        packages = self.ordering_manager.get_packages_of_type([package_type],
-                                                              mask=mask)
-        if len(packages) != 1:
+        package_keyname = 'BARE_METAL_SERVER'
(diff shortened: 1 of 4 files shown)


allmightyspiff:
Can you apply this change to  https://github.com/softlayer/softlayer-python/blob/c7e3d847c6b4533514a0860702a519d6a5004ea6/SoftLayer/managers/vs.py#L817  
so that both hardware and virtual use your get_package_by_key method, then we can remove 

get_package_id_by_type
get_package_by_type
get_packages_of_type
from the ordering manager and clean it up a bit.

kwienken:
@allmightyspiff I've updated the virtual manager to use get_package_by_key but I left the _by_type methods in the ordering manager. After talking with @sergiocarlosmorales I think it's best that we leave them in to prevent breaking other users applications that may include the package and use the ordering library.

sergiocarlosmorales:
Agree here, if we were to remove that functionality we might break existing customer implementations and since it would be backwards incompatible, we'd likely need to bump the release version. Too much, let's keep it simple for now :)

camporter:
There appears to be reports of issues with packages having the same type and the package not being chosen here before this change (on 5.2.1). @allmightyspiff any way we could get a new minor or bugfix release out with this included?

allmightyspiff:
5.2.3 is now released.

PR #816, opened 2017-04-25
Update slcli doc with new subcommands

rlrossiter:
The CLI doc still had some old commands in it, so those have been
removed. There have also been new commands added to slcli that weren't
reflected in the doc.


allmightyspiff:
Looks good to me.

Thanks.

PR #819, opened 2017-04-28
Implementation for volume-counting

kyubifire:
2 new commands:

slcli block volume-count
slcli file volume-count

improved display of information in block and file volume-list


SoftLayer/CLI/file/count.py
```diff
@@ -0,0 +1,55 @@
...
+              help='Columns to display. Options: {0}'.format(
+                  ', '.join(column.name for column in COLUMNS)),
+              default=','.join(DEFAULT_COLUMNS))
+@environment.pass_env
+def cli(env, sortby, columns, datacenter):
+    """List number of file storage volumes per datacenter."""
+
+    file_manager = SoftLayer.FileStorageManager(env.client)
+    file_volumes = file_manager.list_file_volumes(datacenter=datacenter,
+                                                  mask=columns.mask())
```
allmightyspiff: instead of modifying the mask in the manager, just do something like this. fix the line width so tox doesn't complain about it too.

```
    mask = "mask[serviceResource[datacenter[name]],replicationPartners[serviceResource[datacenter[name]]]]"
    block_volumes = block_manager.list_block_volumes(datacenter=datacenter,
                                                     mask=mask)
```

SoftLayer/CLI/block/count.py
```diff
@@ -0,0 +1,55 @@
...
+@click.option('--columns',
+              callback=column_helper.get_formatter(COLUMNS),
+              help='Columns to display. Options: {0}'.format(
+                  ', '.join(column.name for column in COLUMNS)),
+              default=','.join(DEFAULT_COLUMNS))
+@environment.pass_env
+def cli(env, sortby, columns, datacenter):
+    """List number of block storage volumes per datacenter."""
+    block_manager = SoftLayer.BlockStorageManager(env.client)
+    block_volumes = block_manager.list_block_volumes(datacenter=datacenter,
```
allmightyspiff: same comment as in the file volume-count CLI command

dpickle2:
Work from this pull request is continued in #856.

PR #821, opened 2017-05-03
Issues820

allmightyspiff:
Fixes bugs caused by https://softlayer.github.io/release_notes/20170503/
Increased flake8 line length to 120 characters. That should be enough for anyone.


allmightyspiff:
whops, lots of old commits in here for some reason.

PR #824, opened 2017-05-18
IPsec API Support and CLI Interaction

rposky:
Adding IPSEC manager and CLI commands to interact with SoftLayer's IPSEC VPN service. Support is added to list, detail, update and configure tunnel contexts, to add and remove subnet associations and to add, remove and update address translation entries.

Unit testing coverage of 100% has been achieved on all but 1 file 'SoftLayer/CLI/vpn/ipsec/translation/update.py', which is at 94%. Overall project coverage is reported as 83.67%.

The following CLI commands have been added:
- ipsec list
- ipsec detail
- ipsec configure
- ipsec update
- ipsec subnet-add
- ipsec subnet-remove
- ipsec translation-add
- ipsec translation-remove
- ipsec translation-update


SoftLayer/CLI/vpn/ipsec/update.py
```diff
@@ -0,0 +1,102 @@
...
+              help='Remote peer IP address value')
+@click.option('--preshared-key',
+              default=None,
+              help='Preshared key value')
+@click.option('--p1-auth',
+              '--phase1-auth',
+              default=None,
+              type=click.Choice(['MD5', 'SHA1', 'SHA256']),
+              help='Phase 1 authentication value')
+@click.option('--p1-crypto',
```
allmightyspiff: I'd like most of these options to have defaults where possible, mirroring what is in the control portal for defaults would be good.

PR #829, opened 2017-06-06
Update file & block volume-detail commands to show duplicate volume info

dpickle2:
[This pull request is being opened now for review, but it should not be merged in yet as some work is still pending on the backend for duplication features to be fully supported.]


iodine53:
It seems like it'd be a good idea to merge this into #830 maybe?

dpickle2:
I'm happy to merge this into #829 if desired. Should I go ahead and do that?

iodine53:
I'm thinking so; while they aren't strictly dependent on each other, one shouldn't be released without the other, so combining them would make sense.

dpickle2:
I have merged the commit into the feature/volume-duplicate branch, so these changes are now included in #830 .

PR #830, opened 2017-06-06
Add volume-duplicate command for file & block storage

dpickle2:
[This pull request is being opened now for review, but it should not be merged in yet as some work is still pending on the backend for duplication features to be fully supported.]


allmightyspiff:
Did some testing, everything seems to work properly.

Issue #835, opened 2017-06-27
"slcli firewall cancle" failed


MarikoTagawa:
**Please triple-check to make sure that you have properly masked out user credentials like usernames, passwords and API keys before submitting your issue**

### Expected Behavior
Cancel the specified firewall

### Actual Behavior
Got the following error and failed.
------------------------------
$ slcli firewall cancel "vlan:8003"
This action will cancel a firewall from your account. Continue? [y/N]: y
An unexpected error has occured:
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/SoftLayer-5.2.0-py2.7.egg/SoftLayer/CLI/core.py", line 163, in main
    cli.main(**kwargs)
  File "/Library/Python/2.7/site-packages/click/core.py", line 680, in main
    rv = self.invoke(ctx)
  File "/Library/Python/2.7/site-packages/click/core.py", line 1027, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Python/2.7/site-packages/click/core.py", line 1027, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Python/2.7/site-packages/click/core.py", line 873, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Library/Python/2.7/site-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/click/decorators.py", line 63, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/Library/Python/2.7/site-packages/click/core.py", line 508, in invoke
    return callback(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/SoftLayer-5.2.0-py2.7.egg/SoftLayer/CLI/firewall/cancel.py", line 30, in cli
    mgr.cancel_firewall(firewall_id, dedicated=True)
  File "/Library/Python/2.7/site-packages/SoftLayer-5.2.0-py2.7.egg/SoftLayer/managers/firewall.py", line 90, in cancel_firewall
    billing_id = fwl_billing['billingItem']['id']
KeyError: 'billingItem'

Feel free to report this error as it is likely a bug:
    https://github.com/softlayer/softlayer-python/issues
------------------------------

### Environment Information
Operating System: Mac OS X 10.12.3
softlayer-python version (`slcli --version`): slcli (SoftLayer Command-line), version 5.2.0

[...]

PR #836, opened 2017-06-28
Check for billing before attempting to cancel a firewall.

camporter:
This prevents an unexpected error from occurring when billing does not exist for a firewall but cancellation is attempted.

Addresses #835


allmightyspiff:
thanks, looks good to me.

Issue #832, opened 2017-06-21
Upgrade storage ordering


allmightyspiff:
currently ordering storage in SLCLI orders "Endurance / Performance Storage", should be upgraded to order "Storage as a Service" where possible.

1. Might need to consider backward compatability for DCs or Customers who want the v1 offering
2. Add a differentiator in the SLCLI for volumes in v1 and v2 offerings.

allmightyspiff:
issues created from a conversation with @dpickle2

bdumler:
Working on a project for hourly order for STaaS to IMS/SL including an SLCLI.    Will be interested to receive updates when the addition of STaaS ordering is completed for the SLCLI.  To add hourly ordering into the SLCLI would be similar to how the SLCLI is used to order hourly servers.  I am happy to share more detail requirements.  Bill Dumler (bdumler@us.ibm.com)

dpickle2:
Pull request 838 has been merged in. If additional details are needed or questions arise, feel free to reach out to me.

PR #838, opened 2017-07-06
Update file/block ordering to support storage_as_a_service package

dpickle2:
**Includes updates to logic for ordering volumes, replicas,
and snapshot space. The storage_as_a_service package is used
as a default for new volume orders, but customers can specify
older packages for use. The package associated with the
existing primary volume is used in replica and snapshot orders.

**Also includes a refactoring of a number of the unit tests
for File and Block manager functions.

Issue #832 should be addressed by this pull request.


SoftLayer/fixtures/SoftLayer_Network_Storage.py
```diff
@@ -1,4 +1,4 @@
-DUPLICATABLE_VOLUME = {
+SAAS_TEST_VOLUME = {
```
tdurden82: Should this be STAAS?

dpickle2: In this case, SAAS was meant to represent "Storage As A Service". This test volume object is intended for use in tests relating to the Storage As A Service package. If it would help clear up possible misunderstanding, I am happy to add a comment to the source to indicate that SAAS corresponds to the Storage As A Service offering/package.

SoftLayer/managers/block.py
```diff
@@ -305,70 +306,26 @@ def delete_snapshot(self, snapshot_id):
...
+        :param storage_type: 'performance' or 'endurance'
         :param location: Datacenter in which to order iSCSI volume
         :param size: Size of the desired volume, in GB
         :param os_type: OS Type to use for volume alignment, see help for list
         :param iops: Number of IOPs for a "Performance" order
         :param tier_level: Tier level to use for an "Endurance" order
         :param snapshot_size: The size of optional snapshot space,
-        if snapshot space should also be ordered (None if not ordered)
+            if snapshot space should also be ordered (None if not ordered)
+        :param service_offering: Requested offering package to use in the order
```
allmightyspiff: can we enumerate the possible options for service_offering here

SoftLayer/CLI/block/order.py
```diff
@@ -48,40 +48,48 @@
               help='Optional parameter for ordering snapshot '
               'space along with endurance block storage; specifies '
               'the size (in GB) of snapshot space to order')
+@click.option('--service-offering',
+              help='The service offering package to use for placing '
+              'the order [optional, default is \'storage_as_a_service\'',
+              type=click.Choice([
+                  'storage_as_a_service',
+                  'enterprise',
```
allmightyspiff: shouldn't this be endurance instead of enterprise?

SoftLayer/managers/block.py
dpickle2: I will update this. Thanks!

SoftLayer/CLI/block/order.py
dpickle2: I chose "enterprise" to correspond to the name of the SoftLayer_Container_Product_Order_Network_Storage_Enterprise order container (http://developer.softlayer.com/reference/datatypes/SoftLayer_Container_Product_Order_Network_Storage_Enterprise). Calling it "endurance" instead "enterprise" might be more clear within the SLCLI, though. Should I go ahead and update this as well @allmightyspiff ?

SoftLayer/CLI/block/order.py
```diff
@@ -48,40 +48,48 @@
               help='Optional parameter for ordering snapshot '
               'space along with endurance block storage; specifies '
               'the size (in GB) of snapshot space to order')
+@click.option('--service-offering',
+              help='The service offering package to use for placing '
+              'the order [optional, default is \'storage_as_a_service\'',
```
dpickle2: Ah, I missed that. Thank you for pointing that out!

SoftLayer/CLI/block/order.py
```diff
@@ -48,40 +48,48 @@
...
 @environment.pass_env
 def cli(env, storage_type, size, iops, tier, os_type,
-        location, snapshot_size):
+        location, snapshot_size, service_offering):
     """Order a block storage volume."""
     block_manager = SoftLayer.BlockStorageManager(env.client)
     storage_type = storage_type.lower()
 
+    if service_offering is None:
+        service_offering = 'storage_as_a_service'
```
dpickle2: That would be cleaner. I will change this to use a default value.

SoftLayer/CLI/block/order.py
allmightyspiff: ok, since that corresponds with how the datatype is named "enterprise" is fine.

PR #840, opened 2017-07-07
STORAUTO-62 support for setting LUN ID on block volumes

causton81:
This depends on FBLOCK-3 being released, but I would like to address any comments in the meantime so it can be ready to go when FBLOCK-3 is released.

### DIFF
--- a/SoftLayer/CLI/block/lun.py
+++ b/SoftLayer/CLI/block/lun.py
@@ -1,0 +1,36 @@
+"""Set the LUN ID on an iSCSI volume."""
+# :license: MIT, see LICENSE for more details.
+
+import click
+
+import SoftLayer
+from SoftLayer.CLI import environment
+
+
+@click.command()
+@click.argument('volume-id')
+@click.argument('lun-id')
+@environment.pass_env
+def cli(env, volume_id, lun_id):
+    """Set the LUN ID on an existing block storage volume.
+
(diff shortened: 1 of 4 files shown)


SoftLayer/CLI/block/lun.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    block_storage_manager = SoftLayer.BlockStorageManager(env.client)
+
+    res = block_storage_manager.create_or_update_lun_id(volume_id, lun_id)
+
+    if 'value' in res and lun_id == res['value']:
+        click.echo(
+            'Block volume with id %s is reporting LUN ID %s' % (res['volumeId'], res['value']))
+    else:
```
allmightyspiff: how does createOrUpdateLunId report errors?
if it throws an exception, this else could be removed
if it returns false this is fine, but if it returns a message, that message should be displayed.

causton81: There is a bug that returns null when the underlying property does not exist before the method is called. I am told that is addressed in another issue.

allmightyspiff:
@causton81 looks like my change in #841 is causing a merge conflict with this pull request.

When you get a chance can you update your changes to include mine to resolve this conflict.

Thanks.

causton81:
@allmightyspiff , FBLOCK-3 was deployed and this should be good to merge now.

PR #848, opened 2017-07-26
Add support for dedicated host instances to virtual server upgrades

kwienken:
Changing to use SL_Virtual_Guest.getUpgradeItemPrices which returns only the prices available for upgrade. This allows for upgrading without duplicating business logic in the client.


sergiocarlosmorales:
Merged & released.
Thanks for reporting it!

Issue #847, opened 2017-07-26
'slcli hardware detail 11744' generates traceback


briancn:
**Please triple-check to make sure that you have properly masked out user credentials like usernames, passwords and API keys before submitting your issue**

'slcli hardware detail 11744' command causes error/traceback

### Actual Behavior
An unexpected error has occured:
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/SoftLayer-5.2.8-py2.7.egg/SoftLayer/CLI/core.py", line 163, in main
    cli.main(**kwargs)
  File "/Library/Python/2.7/site-packages/click-6.7-py2.7.egg/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Library/Python/2.7/site-packages/click-6.7-py2.7.egg/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Python/2.7/site-packages/click-6.7-py2.7.egg/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Python/2.7/site-packages/click-6.7-py2.7.egg/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Library/Python/2.7/site-packages/click-6.7-py2.7.egg/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/click-6.7-py2.7.egg/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/Library/Python/2.7/site-packages/click-6.7-py2.7.egg/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/SoftLayer-5.2.8-py2.7.egg/SoftLayer/CLI/hardware/detail.py", line 46, in cli
    table.add_row(['memory', formatting.gb(result['memoryCapacity'])])
  File "/Library/Python/2.7/site-packages/SoftLayer-5.2.8-py2.7.egg/SoftLayer/CLI/formatting.py", line 131, in gb
    return FormattedItem(int(float(gigabytes)) * 1024,
TypeError: float() argument must be a string or a number

### Environment Information
Operating System:
Mac 10.12.5 (16F2073)

softlayer-python version (`slcli --version`):
[...]

PR #849, opened 2017-07-27
Avoid blindly passing memory result to formatter

sergiocarlosmorales:


sergiocarlosmorales:
Relates to https://github.com/softlayer/softlayer-python/issues/847

PR #852, opened 2017-07-31
File and Block Hourly billing feature changes for slci

svrajput:
Please merge after Davids pull request to see actual changes (  https://github.com/softlayer/softlayer-python/pull/838 )


svrajput:
Yeah sure I will take care of conflicts and shall update pull request.

tests/managers/file_tests.py
```diff
@@ -166,6 +181,27 @@ def test_cancel_snapshot_immediately(self):
...
+            'billingItem': {
+                'activeChildren': [
+                    {'categoryCode': 'storage_snapshot_space', 'id': 417}
+                ],
+                'hourlyFlag': True,
+                'id': 449
+            },
+        }
+
+        self.file.cancel_snapshot_space(1234, immediate=True)
```
svrajput: For hourly billing we can cancel snapshot immediate, its part of requirement.

tests/managers/block_tests.py
```diff
@@ -181,6 +196,27 @@ def test_cancel_snapshot_immediately(self):
...
+            'billingItem': {
+                'activeChildren': [
+                    {'categoryCode': 'storage_snapshot_space', 'id': 417}
+                ],
+                'hourlyFlag': True,
+                'id': 449
+            },
+        }
+
+        self.block.cancel_snapshot_space(1234, immediate=True)
```
svrajput: Its correct, user can cancel hourly billing instance on immediate basis. its part of requirement.

SoftLayer/CLI/block/order.py
```diff
@@ -56,13 +56,28 @@
                   'storage_as_a_service',
                   'enterprise',
                   'performance']))
+@click.option('--billing',
+              type=click.Choice(['hourly', 'monthly']),
+              default='monthly',
+              show_default=True,
+              help="Optional parameter for Billing rate. Default to monthly")
```
svrajput: show_default=True - removed as per discussion with David. will push latest file in a while.

tests/managers/block_tests.py
dpickle2: Based on the current logic in the Block (and File) managers, immediate will always be used if the volume has hourly billing. Passing in False in the unit test is meant to ensure that the immediate flag is correctly changed to be True when the API call is actually made.

svrajput:
Unstable build .. hence closing. will create new pull request.

PR #856, opened 2017-08-07
Feature/volume-count

dpickle2:
(follow-up for PR #819)
This pull request includes a new branch in which I cleaned up the commit history a bit and addressed the feedback on pull request #819.


dpickle2:
It was not addressed in the previous pull request, but I did notice that the volume-count commands include the optional argument, "--columns". I was not sure if this served any practical purpose, since the output table for this command only has two columns. If desired, I'm happy to remove this optional argument, but I wanted to verify that it was not desired beforehand.

allmightyspiff:
Can I close #819 ?

Since there are only 2 columns removing --columns would make sense to me.

dpickle2:
Alright, I have removed the '--columns' argument from the volume-count commands.

PR #861, opened 2017-08-17
Allow `hardware.place_order` with size ID (and name)

ZuluPro:


allmightyspiff:
The CI check should be fixed in the latest master commit. 

Also, are there any presets that can't be ordered here with just their size name?

kwienken:
@allmightyspiff I do not see anything here that would cause problems with ordering a preset. 

I'm not thrilled with allowing an id to be used instead of the size (keyName) since we tend to encourage keyNames over ids. I guess it's not that much of a risk though since a preset id isn't as likely to change as something like a price id.

ZuluPro:
Hello,@kwienken
In fact, I'm using ID because (default) hardware listing doesn't give the preset's name but its ID.

kwienken:
@ZuluPro in that case I'm fine with using the ID instead of keyName.

SoftLayer/managers/hardware.py
```diff
@@ -467,13 +467,18 @@ def _generate_create_dict(self,
             'hostname': hostname,
             'domain': domain,
         }
-
+        try:
+            size = _get_preset_id(package, size)
+        except SoftLayer.SoftLayerError as err:
+            if not err.message.startswith('Could not find valid size for'):
+                raise
+            size = size
```
ZuluPro: As you asked I check if a size exists with its name before take ID.

I check with try/except, and raise other exceptions.

ZuluPro:
@allmightyspiff @kwienken 
Any news ?

ZuluPro:
Any news ?

allmightyspiff:
I'm going to address this in #912

PR #864, opened 2017-08-23
updating the virt cli create and create-options commands

kwienken:
The virtual create-options command was updated to show flavors and the dedicated host options. The virtual create command was updated to accept a flavor key and a host id. 

This pull request is dependent on some minor updates to the SoftLayer API. DO NOT MERGE. Myself, @sergiocarlosmorales, or @camporter will handle merging this one at the appropriate time.

### DIFF
--- a/SoftLayer/CLI/virt/create.py
+++ b/SoftLayer/CLI/virt/create.py
@@ -25,8 +25,6 @@ def _update_with_like_args(ctx, _, value):
     like_args = {
         'hostname': like_details['hostname'],
         'domain': like_details['domain'],
-        'cpu': like_details['maxCpu'],
-        'memory': '%smb' % like_details['maxMemory'],
         'hourly': like_details['hourlyBillingFlag'],
         'datacenter': like_details['datacenter']['name'],
         'network': like_details['networkComponents'][0]['maxSpeed'],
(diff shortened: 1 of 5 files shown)


SoftLayer/CLI/virt/create.py
```diff
@@ -305,6 +327,18 @@ def cli(env, **args):
...
 
+    if all([args['cpu'], args['flavor']]):
+        raise exceptions.ArgumentError(
+            '[-c | --cpu] not allowed with [-f | --flavor]')
+
+    if all([args['memory'], args['flavor']]):
+        raise exceptions.ArgumentError(
+            '[-m | --memory] not allowed with [-f | --flavor]')
+
+    if all([args['dedicated'], args['flavor']]):
```
camporter: You could also include one for `host-id` and `flavor` right?

SoftLayer/CLI/virt/create_options.py
```diff
@@ -25,35 +25,61 @@ def cli(env):
...
-    ded_cpu = [x for x in result['processors']
-               if x['template'].get('dedicatedAccountHostOnlyFlag',
-                                    False)]
-
-    def add_cpus_row(cpu_options, name):
-        """Add CPU rows to the table."""
-        cpus = []
-        for cpu_option in cpu_options:
-            cpus.append(str(cpu_option['template']['startCpus']))
+    bal_flavors = [str(x['flavor']['keyName']) for x in result['flavors']
```
camporter: You could put flavor extraction into a separate method and generalize it to avoid duplication here.

sergiocarlosmorales: also what does bal mean?

sergiocarlosmorales: +1 seems like a lot of duplication in this area

SoftLayer/managers/vs.py
```diff
@@ -299,38 +300,49 @@ def _generate_create_dict(
...
+            private_security_groups=None, **kwargs):
         """Returns a dict appropriate to pass into Virtual_Guest::createObject
 
             See :func:`create_instance` for a list of available options.
         """
-        required = [cpus, memory, hostname, domain]
+        required = [hostname, domain]
+
+        flavor = kwargs.get('flavor', None)
+        host_id = kwargs.get('host_id', None)
```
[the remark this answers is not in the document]
kwienken: host_id should not be provided with flavor. I'll update to prevent that.

SoftLayer/CLI/virt/create.py
kwienken: Yep. Adding.

SoftLayer/CLI/virt/create_options.py
kwienken: Will do.

PR #879, opened 2017-09-28
resource metadata update 

suppandi:
eg: need to call as getDomain, not Domain.


allmightyspiff:
this should be fixed in the METADATA_MAPPING list, not the in call itself, since that will break all the other calls.

suppandi:
all the functions in http://sldn.softlayer.com/reference/services/SoftLayer_Resource_Metadata are getXXX calls

Issue #885, opened 2017-10-12
createObjects methods do not work with REST API


rlrossiter:
**Please triple-check to make sure that you have properly masked out user credentials like usernames, passwords and API keys before submitting your issue**

### Expected Behavior
createObjects with the RestTransport should operate the same way as the XmlRpcTransport. Calling createObjects with multiple objects should create each object, and return all created objects

### Actual Behavior
Here is a script that I wrote to attempt creating multiple VSIs as the comments in the VS manager's create_instances function instructs. The endpoint I was pointing to was `endpoint_url = https://api.softlayer.com/rest/v3.1/`
```
import os

import SoftLayer

config_file = os.path.expanduser('~/.softlayer')

client = SoftLayer.create_client_from_env(config_file=config_file)
vs = SoftLayer.VSManager(client)

new_vsi = {
    'domain': 'sl.test',
    'hostname': 'sl-test',
    'datacenter': 'dal13',
    'cpus': 1,
    'os_code' : 'UBUNTU_LATEST',
    'hourly': True,
    'disks': ('100',),
    'local_disk': True,
    'memory': 1024
}

instances = [new_vsi.copy(), new_vsi.copy()]
instances[0]['hostname'] = 'sl-test0'
instances[1]['hostname'] = 'sl-test1'

vsis = vs.create_instances(instances)
for vsi in vsis:
    print vsi
```

The error output was:
```
Traceback (most recent call last):
  File "test_createObjects.py", line 26, in <module>
    vsis = vs.create_instances(instances)
  File "/Users/rlrossit/venv/lib/python2.7/site-packages/SoftLayer/managers/vs.py", line 635, in create_instances
    for kwargs in config_list])
  File "/Users/rlrossit/venv/lib/python2.7/site-packages/SoftLayer/API.py", line 392, in call_handler
    return self(name, *args, **kwargs)
  File "/Users/rlrossit/venv/lib/python2.7/site-packages/SoftLayer/API.py", line 360, in call
    return self.client.call(self.name, name, *args, **kwargs)
  File "/Users/rlrossit/venv/lib/python2.7/site-packages/SoftLayer/API.py", line 263, in call
    return self.transport(request)
  File "/Users/rlrossit/venv/lib/python2.7/site-packages/SoftLayer/transports.py", line 315, in __call__
    message)
[...]

PR #886, opened 2017-10-13
Issues885

allmightyspiff:
addresses #885


allmightyspiff:
## Changes
- Fixed an issues with the Rest transport not adding the method to createObject and createObjects calls
- added route to hardware:power:rescue 
- added unit tests to hardware:power:rescue

PR #888, opened 2017-10-22
virtual detail: show dedicated host info, if applicable

flavio-fernandes:
This is only applicable for VSIs that were created to be in a
dedicated host.

Provide id and name of the dedicated host where VSI is running.

### DIFF
--- a/SoftLayer/CLI/virt/detail.py
+++ b/SoftLayer/CLI/virt/detail.py
@@ -45,6 +45,7 @@ def cli(env, identifier, passwords=False, price=False):
     table.add_row(['active_transaction', formatting.active_txn(result)])
     table.add_row(['datacenter',
                    result['datacenter']['name'] or formatting.blank()])
+    _cli_helper_dedicated_host(env, result, table)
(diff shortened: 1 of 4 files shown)


SoftLayer/CLI/virt/detail.py
```diff
@@ -138,3 +139,19 @@ def cli(env, identifier, passwords=False, price=False):
...
+
+    dedicated_host_id = utils.lookup(result, 'dedicatedHost', 'id')
+    if dedicated_host_id:
+        table.add_row(['dedicated_host_id', dedicated_host_id])
+        # Try to find name of dedicated host
+        try:
+            dedicated_host = env.client.call('Virtual_DedicatedHost', 'getObject',
+                                             id=dedicated_host_id)
+        except SoftLayer.SoftLayerAPIError:
+            dedicated_host = {}
```
allmightyspiff: At least log an error when the API throws an exception.

allmightyspiff:
looks good, I'll merge this in the morning.

PR #894, opened 2017-11-21
Remove redundant validation from duplicate ordering (covered by API)

dpickle2:
With the upcoming introduction of volume modification, some of the existing size-validation logic for ordering duplicate volumes is no longer valid in all cases. In this branch, I have removed most of the size and performance validation logic for ordering duplicate volumes, since this validation is done within the API anyway. While increased reliance on the API does limit how specific we can be with the SLCLI error messages, it does seem like reducing redundant validation logic in the SLCLI will help increase the SLCLI's maintainability overall. Please let me know if I have removed too much, and I will be happy to add logic back in if needed.

This branch also cleans up some unit tests by using deepcopy for copying test fixtures. This will prevent an undesirable "cascading test failure" effect that came with my previous implementation of some unit tests.

### DIFF
--- a/SoftLayer/managers/storage_utils.py
+++ b/SoftLayer/managers/storage_utils.py
@@ -887,8 +887,8 @@ def prepare_duplicate_order_object(manager, origin_volume, iops, tier,
 
     :param manager: The File or Block manager calling this function
     :param origin_volume: The origin volume which is being duplicated
-    :param iops: The IOPS per GB for the duplicant volume (performance)
-    :param tier: The tier level for the duplicant volume (endurance)
+    :param iops: The IOPS for the duplicate volume (performance)
+    :param tier: The tier level for the duplicate volume (endurance)
     :param duplicate_size: The requested size for the duplicate volume
     :param duplicate_snapshot_size: The size for the duplicate snapshot space
     :param volume_type: The type of the origin volume ('file' or 'block')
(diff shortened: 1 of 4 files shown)


dpickle2:
It says coverage decreased, but there is less code to test now, so I'm not sure how to fix this one without adding unit tests elsewhere to unrelated code. I'll try to figure out what to do to address this.

SoftLayer/managers/storage_utils.py
```diff
@@ -942,13 +942,14 @@ def prepare_duplicate_order_object(manager, origin_volume, iops, tier,
...
-    if origin_storage_type == 'PERFORMANCE_BLOCK_STORAGE'\
-            or origin_storage_type == 'PERFORMANCE_BLOCK_STORAGE_REPLICANT'\
-            or origin_storage_type == 'PERFORMANCE_FILE_STORAGE'\
-            or origin_storage_type == 'PERFORMANCE_FILE_STORAGE_REPLICANT':
+    if 'PERFORMANCE' in origin_storage_type:
         volume_is_performance = True
-        iops = _validate_dupl_performance_iops(
-            origin_volume, iops, duplicate_size)
+        if iops is None:
+            if isinstance(utils.lookup(origin_volume, 'provisionedIops'), str):
```
allmightyspiff: I think this check is a little unneeded. What if the API changes provisionedIops to be an int in the future?
I think just doing ```iops = int(origin_volume.get('provisionedIops', 0))``` should be sufficient. Then you can make sure that value is > 0 and then throw your exception.

dpickle2: I can do that instead, no problem.

PR #896, opened 2017-11-22
Dedicate command

khuong507:


SoftLayer/managers/dedicated_host.py
```diff
@@ -0,0 +1,333 @@
...
+        for price in package['prices']:
+            if price.get('locationGroupId') is '':
+                return price['id']
+
+        raise SoftLayer.SoftLayerError(
+            "Could not find valid price")
+
+    def _get_item(self, package):
+        """Returns the item for ordering a dedicated host."""
+        description = '56 Cores X 242 RAM X 1.2 TB'
```
allmightyspiff: I can see this being problematic if we ever have more dedicated host sizes. I'd also prefer we check keynames not descriptions.
``` _get_item(self, package, itemKey='DEDICATED_HSOT')```

or something like that.

File: SoftLayer/managers/dedicated_host.py [hunk 2]
```diff
@@ -0,0 +1,333 @@
...
+        mask = '''
+            id,
+            hostname
+        '''
+
+        if locations is not None:
+            for location in locations:
+                if location['locationId'] is not None:
+                    loc_id = location['locationId']
+                    host = {
```
allmightyspiff: I can see this being problematic if more dedicated host sizes are added. This should be calculated from the package or supplied by the CLI I think. 

Unless its possible to not supply this data at all and just get a list of routers without specifying a template

SoftLayer/CLI/dedicatedhost/create.py
```diff
@@ -0,0 +1,99 @@
...
+              prompt=True)
+@click.option('--datacenter', '-d', help="Datacenter shortname",
+              required=True,
+              prompt=True)
+@click.option('--billing',
+              type=click.Choice(['hourly', 'monthly']),
+              default='hourly',
+              show_default=True,
+              help="Billing rate")
+@click.option('--test',
```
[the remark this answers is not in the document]
khuong507: Currently virtual servers and hardware use --test. Would you still like it changed?

SoftLayer/CLI/dedicatedhost/create.py
```diff
@@ -0,0 +1,99 @@
...
+from SoftLayer.CLI import helpers
+
+
+@click.command(
+    epilog="See 'slcli dedicatedhost create-options' for valid options.")
+@click.option('--hostname', '-H',
+              help="Host portion of the FQDN",
+              required=True,
+              prompt=True)
+@click.option('--router', '-r',
```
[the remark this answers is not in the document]
khuong507: 2. Specifying a VLAN is not an option for dedicated hosts

SoftLayer/CLI/hardware/create.py
```diff
@@ -122,6 +122,7 @@ def cli(env, **args):
         return
 
     if do_create:
+
```
rlrossiter: I don't think you wanted to be changing this file

SoftLayer/fixtures/SoftLayer_Product_Package.py
```diff
@@ -892,3 +887,237 @@
         'setupFee': '0',
         'sort': 0
     }]
+
+verifyOrderDH = {
```
rlrossiter: Not sure what the normal process is around adding fixtures, but this seems a little long. Do we want this reduced down so it only has the fields that you need? This question is more for @allmightyspiff

SoftLayer/managers/dedicated_host.py
```diff
@@ -0,0 +1,363 @@
...
+            _filter['dedicatedHosts']['memoryCapacity'] = (
+                utils.query_filter(memory))
+
+        if datacenter:
+            _filter['dedicatedHosts']['datacenter']['name'] = (
+                utils.query_filter(datacenter))
+
+        kwargs['filter'] = _filter.to_dict()
+        func = getattr(self.account, call)
+        return func(**kwargs)
```
rlrossiter: instead of making this roundabout way of calling getDedicatedHosts, I'd just use `self.account.getDedicatedHosts(**kwargs)`

SoftLayer/managers/dedicated_host.py
```diff
@@ -0,0 +1,363 @@
...
+                        'datacenter': {
+                            'id': loc_id
+                        }
+                    }
+                    routers = self.host.getAvailableRouters(host, mask=mask)
+                    return routers
+
+        raise SoftLayer.SoftLayerError("Could not find available routers")
+
+    def _get_default_router(self, routers, router_name):
```
rlrossiter: I think for readability, since router_name isn't required, default it to None in the args here. Even though you always supply it above (and it's None because it's optional in _generate_create_dict()), I'd still be explicit with `router_name=None`

SoftLayer/CLI/dedicatedhost/detail.py
```diff
@@ -0,0 +1,71 @@
...
+    table.add_row(['cpu count', result['cpuCount']])
+    table.add_row(['memory capacity', result['memoryCapacity']])
+    table.add_row(['disk capacity', result['diskCapacity']])
+    table.add_row(['create date', result['createDate']])
+    table.add_row(['modify date', result['modifyDate']])
+    table.add_row(['router id', result['backendRouter']['id']])
+    table.add_row(['router hostname', result['backendRouter']['hostname']])
+    if utils.lookup(result, 'billingItem') != {}:
+        table.add_row(['owner', formatting.FormattedItem(
+            utils.lookup(result, 'billingItem', 'orderItem',
```
rlrossiter: wouldn't this lookup just return {} if billingItem wasn't set, so the if above isn't needed? Because either way if something in that lookup isn't set, you'll end up with `formatting.blank()` anyways

SoftLayer/managers/dedicated_host.py
```diff
@@ -0,0 +1,363 @@
...
+            kwargs['mask'] = (
+                'id,'
+                'name,'
+                'cpuCount,'
+                'memoryCapacity,'
+                'diskCapacity,'
+                'createDate,'
+                'modifyDate,'
+                'backendRouter[id, hostname, domain],'
+                'billingItem[id, nextInvoiceTotalRecurringAmount, '
```
rlrossiter: I'd think about reorganizing this to make it a little more readable, because it's not obvious the nesting structure of this mask (i.e. children and orderItem are under billingItem, but the brackets are hard to follow)

Maybe look at how masking is done where a docstring is used, where different indentation can be used to make nesting easier to read. The only problem with the docstring method is printing it out in the debugger can get a little difficult to read.

khuong507:
Just made the requested changes.

SoftLayer/fixtures/SoftLayer_Product_Package.py
allmightyspiff: I don't really have a strong opinion on this at the moment.

allmightyspiff:
```bash
$ ./slcli  dedicatedhost create -H dedicatedHostTest -D cgallo.com -d  mex01 -f 56_CORES_X_242_RAM_X_1_4_TB --verify
Total hourly cost  0.00
 -- ! Prices reflected here are retail and do not take account level discounts and are not guaranteed.


./slcli  dedicatedhost create -H dedicatedHostTest -D cgallo.com -d  mex01 -f 56_CORES_X_242_RAM_X_1_4_TB
This action will incur charges on your account. Continue? [y/N]: y
Total hourly cost  0.00
 -- ! Prices reflected here are retail and do not take account level discounts and are not guaranteed.
id       21462711
created  2017-12-15T10:54:15-06:00


 ./slcli  dedicatedhost list
50701  ajcb-test          56  1200  242  mex01  1
10201  dedicated2         56  1200  242  dal10  NULL
64801  dedicatedHostTest  56  1200  242  mex01  NULL
10405  fmirDH             56  1200  242  dal09  1

```

PR #899, opened 2017-11-30
Order enhancements

rlrossiter:
This change adds enhancements to be able to use verifyOrder() and placeOrder() with keyNames that come from the API. There are also CLI enhancements to be able to do lookups to find packages, presets, categories, and items from the API, which are then fed into the verify/place_order functions.

### DIFF
--- a/SoftLayer/managers/ordering.py
+++ b/SoftLayer/managers/ordering.py
@@ -6,6 +6,8 @@
     :license: MIT, see LICENSE for more details.
 """
 
+from SoftLayer import exceptions
+
 
 class OrderingManager(object):
     """Manager to help ordering via the SoftLayer API.
@@ -245,7 +247,7 @@ class OrderingManager(object):
 
         package = self.get_package_by_key(package_keyname, mask='id')
         if not package:
-            raise AttributeError("Package {} does not exist".format(package_keyname))
+            raise exceptions.SoftLayerError("Package {} does not exist".format(package_keyname))
 
         categories = self.package_svc.getConfiguration(id=package['id'], **get_kwargs)
         return categories
@@ -269,7 +271,7 @@ class OrderingManager(object):
 
         package = self.get_package_by_key(package_keyname, mask='id')
         if not package:
-            raise AttributeError("Package {} does not exist".format(package_keyname))
+            raise exceptions.SoftLayerError("Package {} does not exist".format(package_keyname))
 
         items = self.package_svc.getItems(id=package['id'], **get_kwargs)
         return items
@@ -315,10 +317,12 @@ class OrderingManager(object):
 
         package = self.get_package_by_key(package_keyname, mask='id')
         if not package:
-            raise AttributeError("Package {} does not exist".format(package_keyname))
+            raise exceptions.SoftLayerError("Package {} does not exist".format(package_keyname))
 
-        presets = self.package_svc.getActivePresets(id=package['id'], **get_kwargs)
-        return presets
+        acc_presets = self.package_svc.getAccountRestrictedActivePresets(
+            id=package['id'], **get_kwargs)
+        active_presets = self.package_svc.getActivePresets(id=package['id'], **get_kwargs)
+        return acc_presets + active_presets
 
     def get_preset_by_key(self, package_keyname, preset_keyname, mask=None):
         """Gets a single preset with the given key."""
@@ -328,7 +332,7 @@ class OrderingManager(object):
         presets = self.list_presets(package_keyname, mask=mask, filter=_filter)
 
         if len(presets) == 0:
-            raise AttributeError(
+            raise exceptions.SoftLayerError(
                 "Preset {} does not exist in package {}".format(preset_keyname,
                                                                 package_keyname))
 
@@ -349,7 +353,7 @@ class OrderingManager(object):
         """
         package = self.get_package_by_key(package_keyname, mask='id')
         if not package:
-            raise AttributeError("Package {} does not exist".format(package_keyname))
+            raise exceptions.SoftLayerError("Package {} does not exist".format(package_keyname))
 
         mask = 'id, keyName, prices'
         items = self.list_items(package_keyname, mask=mask)
@@ -362,7 +366,7 @@ class OrderingManager(object):
                 matching_item = [i for i in items
                                  if i['keyName'] == item_keyname][0]
             except IndexError:
-                raise AttributeError(
+                raise exceptions.SoftLayerError(
                     "Item {} does not exist for package {}".format(item_keyname,
                                                                    package_keyname))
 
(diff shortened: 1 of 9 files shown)


SoftLayer/managers/ordering.py
```diff
@@ -209,15 +213,250 @@ def get_package_by_key(self, package_keyname, mask=None):
...
+
+        acc_presets = self.package_svc.getAccountRestrictedActivePresets(
+            id=package['id'], **get_kwargs)
+        active_presets = self.package_svc.getActivePresets(id=package['id'], **get_kwargs)
+        return acc_presets + active_presets
+
+    def get_preset_by_key(self, package_keyname, preset_keyname, mask=None):
+        """Gets a single preset with the given key."""
+        preset_operation = '_= %s' % preset_keyname
+        _filter = {'activePresets': {'keyName': {'operation': preset_operation}}}
```
rlrossiter: This is only looking in the active presets. It also needs to look at the account restricted active presets with this filter to possibly find it in there as well

rlrossiter: turns out this filter works with both activePresets and accountRestrictedActivePresets, so nothing needs to be updated.

allmightyspiff:
```bash
$ ./slcli --really order place   --complex-type SoftLayer_Container_Product_Order_Hardware_Server --preset D2690_128GB_2X600GBSAS_RAID1_2  --extras '{"hardware":[{"hostname":"testORdering", "domain": "cgallo.com"}]}' BARE_METAL_SERVER TORONTO  1_IP_ADDRESS UNLIMITED_SSL_VPN_USERS_1_PPTP_VPN_USER_PER_ACCOUNT REBOOT_KVM_OVER_IP OS_UBUNTU_16_04_LTS_XENIAL_XERUS_64_BIT BANDWIDTH_0_GB_2 100_MBPS_PRIVATE_NETWORK_UPLINK
id       21469963
created  2017-12-15T15:41:31-06:00
status   PENDING_AUTO_APPROVAL
```

Noticed some oddness with locations though, I'm not sure why DAL13 is an invalid location here. Using any of the 3 letter dc names doens't seem to work.
```
$ ./slcli --really order place   --complex-type SoftLayer_Container_Product_Order_Hardware_Server --preset D2690_128GB_2X600GBSAS_RAID1_2  --extras '{"hardware":[{"hostname":"testORdering", "domain": "cgallo.com"}]}' BARE_METAL_SERVER DAL13 1_IP_ADDRESS UNLIMITED_SSL_VPN_USERS_1_PPTP_VPN_USER_PER_ACCOUNT REBOOT_KVM_OVER_IP OS_UBUNTU_16_04_LTS_XENIAL_XERUS_64_BIT BANDWIDTH_0_GB_2 100_MBPS_PRIVATE_NETWORK_UPLINK
SoftLayerAPIError(SoftLayer_Exception_Order_InvalidLocation): The location provided for this order is invalid.
```

Using its keyname seems to work though. Adding in a way to get the available DCs for a package might be nice. 
```
$ ./slcli --really order place   --complex-type SoftLayer_Container_Product_Order_Hardware_Server --preset D2690_128GB_2X600GBSAS_RAID1_2  --extras '{"hardware":[{"hostname":"testORdering", "domain": "cgallo.com"}]}' BARE_METAL_SERVER DALLAS13 1_IP_ADDRESS UNLIMITED_SSL_VPN_USERS_1_PPTP_VPN_USER_PER_ACCOUNT REBOOT_KVM_OVER_IP OS_UBUNTU_16_04_LTS_XENIAL_XERUS_64_BIT BANDWIDTH_0_GB_2 100_MBPS_PRIVATE_NETWORK_UPLINK
SoftLayerAPIError(SoftLayer_Exception_Order_HardwareUnavailable): There is currently no hardware available for package # 200 with preset configuration # 101 in Dallas 13. Please select a different datacenter
```

SoftLayer/order_utils.py
```diff
@@ -0,0 +1,145 @@
+"""Utility functions for mapping packages and categories to complex types."""
+
+
+def get_package_type_map():
```
allmightyspiff: would it make sense to add the virtual server and bare metal server types here?

SoftLayer/order_utils.py
```diff
@@ -0,0 +1,145 @@
...
+        'BARE_METAL_GATEWAY':
+            'SoftLayer_Container_Product_Order_Hardware_Server_Gateway_Appliance',
+        'GATEWAY_RESOURCE_GROUP':
+            'SoftLayer_Container_Product_Order_Gateway_Appliance_Cluster',
+        'HARDWARE_SECURITY_MODULE':
+            'SoftLayer_Container_Product_Order_Hardware_Security_Module',
+    }
+
+
+def get_category_type_map():
```
allmightyspiff: I'm not sure how I feel about this level of hard coding categories. 
Having the user figure out the complexType on their own might be preferable here. Thoughts?

rlrossiter:
>Noticed some oddness with locations though, I'm not sure why DAL13 is an invalid location here. Using any of the 3 letter dc names doens't seem to work.

So the order API requires you to specify the full name for orders when you provide a location. I was thinking of doing a follow-up PR later to allow you to specify the short names for DCs as well.

SoftLayer/order_utils.py
rlrossiter: Apparently what I've been told, we can't map directly from that package to that order type

SoftLayer/order_utils.py
rlrossiter: Yeah, my goal was to just make this as simple as possible for the user, because it's not obviously clear where to find the SoftLayer_Container_Product_Order type that needs to be used. Unfortunately, this was the only way to do that (there is no API call to properly get the type).

Maybe I should just move this out into a different PR that we can debate on (it'll thankfully make this change a little bit smaller). I'll make complexType a kwarg and CLI option so one day we'll be able to make it optional, but for now I'll require it to be set. How does that sound?

allmightyspiff: Yeah thats a good plan.

SoftLayer/managers/ordering.py
```diff
@@ -209,15 +241,289 @@ def get_package_by_key(self, package_keyname, mask=None):
...
+                            Example: A VSI order requires hostname and domain to be set, so
+                            extras will look like the following:
+                                {'virtualGuests': [{'hostname': 'test',
+                                                    'domain': 'softlayer.com'}]}
+        :param int quantity: The number of resources to order
+
+        """
+        # verify the order, and if the order is valid, the proper prices will be filled
+        # into the order template, so we can just send that to placeOrder to order it
+        verified_order = self.verify_order(package_keyname, location, item_keynames,
```
rlrossiter: verifyOrder() can strip data out of the original order in the returned order, so calling placeOrder with the verified order will be omitting some of the original order data. verify_order() and place_order() should then have a same base prepare_order() function that it then either calls verify or place on, depending on the function it is in.

rlrossiter:
@allmightyspiff I think this is now in a final working state if you want to take a look at it. Thanks!

PR #901, opened 2017-12-05
Added option of Interval Scheduling

fryxxie:


iodine53:
So it looks like there's no SLCLI snapshot schedule show command. :| Not _exactly_ in the scope of this issue, I'll raise a new one.

PR #915, opened 2017-12-23
Usage of requests.Session

ZuluPro:
I just experienced the following error:

```
File "/usr/local/lib/python2.7/dist-packages/SoftLayer/API.py", line 360, in call
    return self.client.call(self.name, name, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/SoftLayer/API.py", line 263, in call
    return self.transport(request)
  File "/usr/local/lib/python2.7/dist-packages/SoftLayer/transports.py", line 199, in __call__
    raise exceptions.TransportError(0, str(ex))
SoftLayer.exceptions.TransportError: TransportError(0): ('Connection broken: error("(104, \'ECONNRESET\')",)', error("(104, 'ECONNRESET')",))
```

So I though about the change the transport implementation to allow retries in case connection error.

This also create memoized session.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -29,3 +29,4 @@ Swapnil Khanapurkar <swapnil_khanapurkar@persistent.co.in>
 The SoftLayer Developer Network <sldn@softlayer.com>
 Tim Ariyeh <tim.ariyeh@gmail.com>
 Wissam Elriachy <wissam.elriachy@gmail.com>
+Anthony Monthe (ZuluPro) <anthony.monthe@gmail.com>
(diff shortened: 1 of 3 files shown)


allmightyspiff:
I'm a little hesitant to make a library wide retry right now. I do have a @retry decorator that I've been adding where I feel it is safe. 

I will review this though to see if using the requests retry function works well. Mostly I'm concerned that this might lead to placeOrder calls putting in multiple orders in certain cases.

ZuluPro:
@allmightyspiff 
Actually the retries are implemented only on ConnectionError.
That means it will not retry when fail to read response or with bad status code,
Only TCP Error, before HTTP ones.

allmightyspiff:
I'll merge this in #916

Issue #588, opened 2015-07-27
Request: implement wait_for_ready for bare metal servers


vincentwoo:
You can issue reload requests to bare metal servers via this client library, but there is not currently a sanctioned way for you to wait until this is completed. This is especially important given that reloads take a half hour or more.

PR #917, opened 2018-01-05
Issues588 - hardware wait for ready

allmightyspiff:
adds wait for ready to HW
adds slcli hw ready command
refactors VS wait_for_Ready to have less code repitition


allmightyspiff:
Resolves #588

Issue #921, opened 2018-01-10
No way to create a VM in a specific subnet, only a particular VLAN


fraz3alpha:
When ordering virtual machines through the Softlayer web UI portal you are asked for the target VLAN, then presented with the subnet on which to place the instance. This softlayer library allows you to pass through a VLAN id only, which means when more than one subnet is configured under a VLAN the host will be created in an arbitrary subnet (it might follow a rule, such as a lowest subnet id, or first alphabetically, but I do not know).

### Expected Behavior
It should be possible to request a new VM on a specific subnet on a given VLAN

### Actual Behavior
It is only possible to specify a VLAN, and the VM may be on any of the attached subnets

### Environment Information
Operating System: Mac OSX High Sierra
slcli (SoftLayer Command-line), version 5.2.11

**Abridged Fix**
I was able to find the mechanism to provide this data in a request after a conversation with Softlayer support. For example if suitable parameters are passed in to [_generate_create_dict ](https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/managers/vs.py#L303) we can modify the way the data is constructed for the vlan, e.g.
https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/managers/vs.py#L369:
```
         if private_vlan:
 -            data.update({
 -                "primaryBackendNetworkComponent": {
 -                    "networkVlan": {"id": int(private_vlan)}}})
 +            # As per https://stackoverflow.com/questions/37592080/create-a-softlayer-virtual-guest-on-a-specific-subnet
 +            # we can specify a subnet under the VLAN
 +            if private_vlan_subnet:
 +                data.update({
 +                    "primaryBackendNetworkComponent": {
 +                        "networkVlanId":int(private_vlan),
 +                        "networkVlan":{
 +                            "primarySubnet":{
 +                                "id":int(private_vlan_subnet)
 +                            }
 +                       }
 +                    }
 +                })
 +            else:
 +                data.update({
 +                    "primaryBackendNetworkComponent": {
 +                        "networkVlan": {"id": int(private_vlan)}}})
```

I have an PR which I can associated with the other supporting code to feed in the `private_vlan_subnet`variable. I will associate this shortly

I spoke directly with @allmightyspiff about this and it was suggested that any enhancement here should also include specifying the subnet on both private and public networks.

PR #923, opened 2018-01-10
WIP: Add support for specifying a subnet on a private VLAN

fraz3alpha:
- For https://github.com/softlayer/softlayer-python/issues/921
  - When a private subnet id is given in addition to a private
    VLAN id this extra information will be passed to the create
    request to further specify the location of the VM
  - TODO: Support specifying subnet on the public network interface
  - TODO: Pass lint checks

### DIFF
--- a/SoftLayer/CLI/virt/create.py
+++ b/SoftLayer/CLI/virt/create.py
@@ -132,6 +132,9 @@ def _parse_create_args(client, args):
     if args.get('vlan_private'):
         data['private_vlan'] = args['vlan_private']
 
+    if args.get('vlan_private_subnet'):
+        data['private_vlan_subnet'] = args['vlan_private_subnet']
+
     if args.get('public_security_group'):
         pub_groups = args.get('public_security_group')
         data['public_security_groups'] = [group for group in pub_groups]
@@ -225,6 +228,10 @@ def _parse_create_args(client, args):
               type=click.INT)
 @click.option('--vlan-private',
               help="The ID of the private VLAN on which you want the virtual "
+                   "server placed",
+              type=click.INT)
+@click.option('--vlan-private-subnet',
+              help="The ID of the private VLAN subnet on which you want the virtual "
                    "server placed",
               type=click.INT)
 @helpers.multi_option('--public-security-group',
--- a/SoftLayer/managers/vs.py
+++ b/SoftLayer/managers/vs.py
@@ -309,7 +309,7 @@ class VSManager(utils.IdentifierMixin, object):
             dedicated=False, public_vlan=None, private_vlan=None,
             userdata=None, nic_speed=None, disks=None, post_uri=None,
             private=False, ssh_keys=None, public_security_groups=None,
-            private_security_groups=None, **kwargs):
+            private_security_groups=None, private_vlan_subnet=None, **kwargs):
         """Returns a dict appropriate to pass into Virtual_Guest::createObject
 
             See :func:`create_instance` for a list of available options.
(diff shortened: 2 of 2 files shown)


fraz3alpha:
Current code fails the tox analysis:
```
R:303, 4: Too many branches (22/20) (too-many-branches)
```

SoftLayer/managers/vs.py
```diff
@@ -368,10 +368,25 @@ def _generate_create_dict(
             data.update({
                 'primaryNetworkComponent': {
                     "networkVlan": {"id": int(public_vlan)}}})
+
         if private_vlan:
```
allmightyspiff: My thoughts here are to break out the private/public vlan stuff into a helper function.


```python
if private_vlan or public_vlan:
    network_component = self._create_network_components(private_vlan, public_vlan, private_vlan_subnet, public_vlan_subnet)
    date.update(network_component)
```

Or something similar would work.

SoftLayer/CLI/virt/create.py
```diff
@@ -227,6 +230,10 @@ def _parse_create_args(client, args):
               help="The ID of the private VLAN on which you want the virtual "
                    "server placed",
               type=click.INT)
+@click.option('--vlan-private-subnet',
+              help="The ID of the private VLAN subnet on which you want the virtual "
+                   "server placed",
```
allmightyspiff: Max line length is 120, so put this all on one line please.

SoftLayer/managers/vs.py
fraz3alpha: Thanks for the suggestions, I had just started doing that very thing - however, just to pre-warn you it seems there is a tox check for max number of lines in a module, and we have just crossed the 1000 limit

allmightyspiff: for that one, you can just add it to the ignored list at the top of this module. I don't want to split this file out at the moment, although I will likely do so in the future.

SoftLayer/CLI/virt/create.py
```diff
@@ -132,6 +132,9 @@ def _parse_create_args(client, args):
     if args.get('vlan_private'):
         data['private_vlan'] = args['vlan_private']
 
+    if args.get('vlan_private_subnet'):
+        data['private_vlan_subnet'] = args['vlan_private_subnet']
```
kwienken: I would recommend naming this private_subnet instead of private_vlan_subnet.

SoftLayer/CLI/virt/create.py
```diff
@@ -227,6 +230,10 @@ def _parse_create_args(client, args):
               help="The ID of the private VLAN on which you want the virtual "
                    "server placed",
               type=click.INT)
+@click.option('--vlan-private-subnet',
```
kwienken: Same here just, --private-subnet.

SoftLayer/managers/vs.py
```diff
@@ -368,10 +368,25 @@ def _generate_create_dict(
...
-            data.update({
-                "primaryBackendNetworkComponent": {
-                    "networkVlan": {"id": int(private_vlan)}}})
+            # As per https://stackoverflow.com/questions/37592080/create-a-softlayer-virtual-guest-on-a-specific-subnet
+            # we can specify a subnet under the VLAN
+            if private_vlan_subnet:
```
kwienken: This shouldn't be inside the private_vlan if statement since I don't think the API requires a vlan id when supplying a subnet id.

fraz3alpha:
Closing this as it was implemented as part of #995 , and that seems to be working well for me now 👍

Issue #932, opened 2018-01-24
edit-securitygroup-rules cannot clear out field


bvanhav:
**Please triple-check to make sure that you have properly masked out user credentials like usernames, passwords and API keys before submitting your issue**

### Expected Behavior
When editing a security group rule using the SLDN API, a string field (for example remote_ip) can be cleared by passing in an empty string. If an empty string is passed to the network manager edit_securitygroup_rule, that empty string should be passed to SLDN and the field cleared.

### Actual Behavior
The empty string is not passed to SLDN and the 'edited' rule returns with the 'cleared' field still set to its original value.

### Environment Information
Operating System:
softlayer-python version (`slcli --version`): 5.2.14

allmightyspiff:
This fix has been merged. Should be included in 5.4.1

PR #933, opened 2018-01-24
Add fix for unsetting of values in edit SG rules

rlrossiter:
The SL API for editing security group rules allows you to unset string
values by passing in an empty string (''). The edit_securitygroup_rule()
function skipped over any values that were falsy (because if they
weren't set in the arguments, we don't want to change them).
Unfortunately, this caused values that were an empty string to also get
skipped over and not get unset in the API. So edit needs to check
explicitly for None (don't change) in order to get skipped over in
passing to the API.

### DIFF
--- a/SoftLayer/managers/network.py
+++ b/SoftLayer/managers/network.py
@@ -318,19 +318,19 @@ class NetworkManager(object):
         """
         successful = False
         obj = {}
-        if remote_ip:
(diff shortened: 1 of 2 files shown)


rlrossiter:
#932

bvanhav:
Looks like this change successfully addresses my concern in #932

PR #936, opened 2018-02-05
Add boot mode option

kwienken:
Add the ability to specify the boot mode when ordering a virtual server.

### DIFF
--- a/SoftLayer/CLI/virt/create.py
+++ b/SoftLayer/CLI/virt/create.py
@@ -177,7 +177,7 @@ def _parse_create_args(client, args):
 @click.option('--image',
               help="Image ID. See: 'slcli image list' for reference")
 @click.option('--boot-mode',
-              help="Specify the mode to boot the OS in",
+              help="Specify the mode to boot the OS in. Supported modes are HVM and PV.",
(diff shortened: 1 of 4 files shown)


SoftLayer/CLI/virt/create.py
```diff
@@ -175,6 +176,9 @@ def _parse_create_args(client, args):
               help="OS install code. Tip: you can specify <OS>_LATEST")
 @click.option('--image',
               help="Image ID. See: 'slcli image list' for reference")
+@click.option('--boot-mode',
+              help="Specify the mode to boot the OS in",
```
allmightyspiff: some context here on what boot modes are available would be good. 
Updating https://softlayer-python.readthedocs.io/en/latest/cli/vs.html (https://github.com/softlayer/softlayer-python/blob/master/docs/cli/vs.rst ) on what choosing a different boot mode does would also be helpful

kwienken: I'll add the boot modes here and get with offering for documentation to add to https://softlayer-python.readthedocs.io/en/latest/cli/vs.html.

PR #956, opened 2018-03-23
Issues922

allmightyspiff:
modifies managers.ordering.generate_order so that it uses a location ID instead of a long name. Can be either short name (dal13) or keyname (DALLAS13). 

Also, default location price id seems to be None now instead of ''

Testing to make sure everything still works.
```
$ ./slcli --really order place --verify --preset D2620V4_64GB_2X1TB_SATA_RAID_1 BARE_METAL_SERVER  TORONTO      OS_UBUNTU_16_04_LTS_XENIAL_XERUS_64_BIT     BANDWIDTH_0_GB_2      1_GBPS_PUBLIC_PRIVATE_NETWORK_UPLINKS      REBOOT_KVM_OVER_IP 1_IP_ADDRESS      UNLIMITED_SSL_VPN_USERS_1_PPTP_VPN_USER_PER_ACCOUNT     --extras '{"hardware": [{"hostname" : "testOrder1", "domain": "cgallo.com"}], "sshKeys" : [87634], "tags": "cgallo, test"}'      --complex-type SoftLayer_Container_Product_Order_Hardware_Server
OS_UBUNTU_16_04_LTS_XENIAL_XERUS_64_BIT              Ubuntu Linux 16.04 LTS Xenial Xerus (64 bit)           0
BANDWIDTH_0_GB_2                                     0 GB Bandwidth                                         0
1_GBPS_PUBLIC_PRIVATE_NETWORK_UPLINKS                1 Gbps Public & Private Network Uplinks                0
REBOOT_KVM_OVER_IP                                   Reboot / KVM over IP                                   0
1_IP_ADDRESS                                         1 IP Address                                           0
UNLIMITED_SSL_VPN_USERS_1_PPTP_VPN_USER_PER_ACCOUNT  Unlimited SSL VPN Users & 1 PPTP VPN User per account  0
RAM_64_GB_DDR3_1333_REG_2                            64 GB RAM                                              0
HARD_DRIVE_1_00_TB_SATA_2                            1.00 TB SATA                                           0
HARD_DRIVE_1_00_TB_SATA_2                            1.00 TB SATA                                           0
REDUNDANT_POWER_SUPPLY                               Redundant Power Supply                                 0
INTEL_INTEL_XEON_E52620_V4_2_10                      Dual Intel Xeon E5-2620 v4 (16 Cores, 2.10 GHz)        0
DISK_CONTROLLER_RAID                                 RAID                                                   0
(py36)
```
```
[...]

PR #958, opened 2018-03-28
Fixed list index out of range error

nanjj:
The error found when `locationGroupId` is None.


allmightyspiff:
locationGroupId needs a check to be None (or possibly '' on some occasions, I need to do some more testing to figure that one out), and not just the first result. 

Thanks for the pull request, but I'm going to fix this in #956, hopefully by March 29.

Issue #959, opened 2018-03-28
Warning/Error messages for ordering File/Block storage using the older service offereings


dpickle2:
It would be helpful for the SLCLI to yield warning or error messages when ordering File/Block storage using the older service offerings (Performance or Endurance/Enterprise, rather than Storage As A Service), since the older service offerings do not allow for various new features which many customers will want (for example, volume duplication and volume modification).

### Expected Behavior
The `slcli file volume-order` and `slcli block volume-order` commands should give a warning message (with prompt for confirmation, perhaps) or simply throw an error (or maybe just remove this functionality, rather than throwing an error?) if an attempt is made to order a block or file volume using the older service offerings (`--service-offering enterprise` or `--service-offering performance`).

### Actual Behavior
No distinction is made (via warning or error) when a block or file volume is ordered with `--service-offering enterprise` or `--service-offering performance`.

PR #969, opened 2018-04-23
#959 added warning message when ordering legacy storage offerings

allmightyspiff:
added warning message when ordering legacy storage offerings from the cli, removed restriction on IOPS being multiples of 100 for performance volumes

### DIFF
--- a/SoftLayer/CLI/block/order.py
+++ b/SoftLayer/CLI/block/order.py
@@ -60,9 +60,9 @@ CONTEXT_SETTINGS = {'token_normalize_func': lambda x: x.upper()}
 @environment.pass_env
 def cli(env, storage_type, size, iops, tier, os_type,
         location, snapshot_size, service_offering, billing):
-    """Order a block storage volume. 
+    """Order a block storage volume.
 
-    Valid size and iops options can be found here: 
+    Valid size and iops options can be found here:
     https://console.bluemix.net/docs/infrastructure/BlockStorage/index.html#provisioning
     """
     block_manager = SoftLayer.BlockStorageManager(env.client)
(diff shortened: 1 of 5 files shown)


SoftLayer/CLI/file/order.py
```diff
@@ -51,30 +49,30 @@
...
+        click.secho('{} is a legacy storage offering'.format(service_offering), fg='red')
+        if hourly_billing_flag:
+            raise exceptions.CLIAbort(
+                'Hourly billing is only available for the storage_as_a_service service offering'
+            )
 
     if storage_type == 'performance':
         if iops is None:
             raise exceptions.CLIAbort(
                 'Option --iops required with Performance')
```
dpickle2: This is a nit-pick detail, but I did notice the similar portion in SoftLayer/CLI/block/order.py (line 84) was modified so that only one line was used for this exception.

SoftLayer/CLI/file/order.py
```diff
@@ -51,30 +49,30 @@
...
 
-        if iops % 100 != 0:
-            raise exceptions.CLIAbort(
-                'Option --iops must be a multiple of 100'
-            )
-
         if service_offering == 'performance' and snapshot_size is not None:
             raise exceptions.CLIAbort(
```
dpickle2: As with my comment on line 74, I don't think this is a major issue, but I noticed the corresponding exception in SoftLayer/CLI/block/order.py (lines 88 and 89) was updated. If possible, it might be nice to update this one as well for consistency.

SoftLayer/CLI/block/order.py
```diff
@@ -49,8 +46,8 @@
               'space along with endurance block storage; specifies '
               'the size (in GB) of snapshot space to order')
 @click.option('--service-offering',
-              help='The service offering package to use for placing '
-              'the order [optional, default is \'storage_as_a_service\']',
+              help="""The service offering package to use for placing the order.
+[optional, default is \'storage_as_a_service\']. enterprise and performance are depreciated""",
```
dpickle2: I think this looks good, but I noticed it is only here in the block volume order command. Would it be possible to add it to SoftLayer/CLI/file/order.py as well?

Issue #975, opened 2018-05-24
Get coverage for CLI/virt/power to 100%


allmightyspiff:
$ tox -e coverage

|File| Coverage| Missing Lines|
| --- | --- | ---|
|SoftLayer/CLI/virt/power.py  | 43% | 19-25, 37-50, 60-71, 80-82, 91-99, 108-110 |

To run just a specific test
$ pytest tests/CLI/modules/vs_tests.py

To run all tests
$ tox

PR #976, opened 2018-05-27
Finished unit tests for CLI/virt/power

erick-sapp:
I believe this completes the unit test / issue #975. If not, let me know if time is available.

### DIFF
--- a/tests/CLI/helper_tests.py
+++ b/tests/CLI/helper_tests.py
@@ -252,21 +252,15 @@ class CLIAbortTests(testing.TestCase):
 class ResolveIdTests(testing.TestCase):
 
     def test_resolve_id_one(self):
-        def resolver(r):
-            return [12345]
-        self.assertEqual(helpers.resolve_id(resolver, 'test'), 12345)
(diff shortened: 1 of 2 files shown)


tests/CLI/helper_tests.py
```diff
@@ -252,16 +252,19 @@ def test_init(self):
 class ResolveIdTests(testing.TestCase):
 
     def test_resolve_id_one(self):
-        resolver = lambda r: [12345]
```
allmightyspiff: Any specific reason for changing this from a lambda function?

erick-sapp: No _real_ reason. The analysis was failing because of it though. Checking it now, it would have been better to skip the assignment and put the lambda expression as the argument itself. That is normally how the lambda is intended to be used.

erick-sapp: I have replace the code with lambdas again, but removed the assignment process it did not like. They are true anonymous functions created within the function calls.

erick-sapp: But in general, I just didn't want the analysis failing for another section of code.

allmightyspiff:
Looks good to me.
This resolves #975

Issue #880, opened 2017-10-02
exception error when cancelling block storage via CLI


engrmg:
**Please triple-check to make sure that you have properly masked out user credentials like usernames, passwords and API keys before submitting your issue**

SLCLI command:

slcli block volume-cancel --immediate <volume id>

### Expected Behavior
Some sort of response acknowledging the cancelling request.

### Actual Behavior
Traceback (most recent call last):

  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/SoftLayer/CLI/core.py", line 163, in main
    cli.main(**kwargs)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/SoftLayer/CLI/block/cancel.py", line 29, in cli
    reason, immediate)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/SoftLayer/managers/block.py", line 472, in cancel_block_volume
    billing_item_id = block_volume['billingItem']['id']
KeyError: 'billingItem'

### Environment Information
Operating System: mac os
[...]

PR #982, opened 2018-06-19
Fixed cancel block storage issue

FernandoOjeda:
Cancel block storage issue https://github.com/softlayer/softlayer-python/issues/880

### DIFF
--- a/SoftLayer/managers/block.py
+++ b/SoftLayer/managers/block.py
@@ -510,20 +510,20 @@ class BlockStorageManager(utils.IdentifierMixin, object):
         block_volume = self.get_block_volume_details(
             volume_id,
             mask='mask[id,billingItem[id,hourlyFlag]]')
-        billingItem = 'billingItem' in block_volume
-        if billingItem is False:
-            print('The block storage was cancelled')
-        else:
-            billing_item_id = block_volume['billingItem']['id']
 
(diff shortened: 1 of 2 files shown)


SoftLayer/managers/block.py
```diff
@@ -510,16 +510,20 @@ def cancel_block_volume(self, volume_id,
         block_volume = self.get_block_volume_details(
             volume_id,
             mask='mask[id,billingItem[id,hourlyFlag]]')
-        billing_item_id = block_volume['billingItem']['id']
+        billingItem = 'billingItem' in block_volume
+        if billingItem is False:
+            print('The block storage was cancelled')
```
allmightyspiff: The managers should not print out anything. I'd rather we raise an exception here.
https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/managers/hardware.py#L87 has a good example of how to handle this sort of logic.

SoftLayer/managers/block.py
```diff
@@ -510,16 +510,20 @@ def cancel_block_volume(self, volume_id,
...
+        else:
```
allmightyspiff: Since were going to throw an exception if we don't have a billing item, the else bits are not really needed and can be reverted back.

Issue #981, opened 2018-06-19
Error in slcli credentials call


npothen:
**Please triple-check to make sure that you have properly masked out user credentials like usernames, passwords and API keys before submitting your issue**

### Expected Behavior
password to be displayed

### Actual Behavior

$ slcli hardware list
:.........:.............:............:................:............:........:
:    id   :   hostname  : primary_ip :   backend_ip   : datacenter : action :
:.........:.............:............:................:............:........:
: 1537051 : ***   :     -   :  -.-.-.-:   dal10    :   -    :
: 1012647 : *** : -.-.-.- :  -.-.-.-:   hou02    :   -    :
: 1000041 : *** : -.-.-.- :  -.-.-.-:   hou02    :   -    :
:  857951 :  *** : -.-.-.- :  -.-.-.-:   hou02    :   -    :
:.........:.............:............:................:............:........:

$ slcli hardware credentials 857951
:..........:..........:
: username : password :
:..........:..........:
:   *****   : ******** :
:..........:..........:

$ slcli hardware credentials 1537051
An unexpected error has occured:
Traceback (most recent call last):
  File "...../lib/python2.7/site-packages/SoftLayer/CLI/core.py", line 163, in main
    cli.main(**kwargs)
  File ".....lib/python2.7/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File ".....lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "...../lib/python2.7/site-packages/click/core.py", line 1060, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "...../lib/python2.7/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File ".....slenv/lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "...../lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "...../lib/python2.7/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "...../lib/python2.7/site-packages/SoftLayer/CLI/hardware/credentials.py", line 25, in cli
    for item in instance['operatingSystem']['passwords']:
KeyError: 'passwords'

Feel free to report this error as it is likely a bug:
[...]

PR #987, opened 2018-06-27
Fixed hardware credentials issue.

FernandoOjeda:
Fixed hardware credentials issue https://github.com/softlayer/softlayer-python/issues/981

### DIFF
--- a/SoftLayer/CLI/hardware/credentials.py
+++ b/SoftLayer/CLI/hardware/credentials.py
@@ -5,9 +5,9 @@ import click
 
 import SoftLayer
 from SoftLayer.CLI import environment
-from SoftLayer import exceptions
 from SoftLayer.CLI import formatting
 from SoftLayer.CLI import helpers
+from SoftLayer import exceptions
 
 
 @click.command()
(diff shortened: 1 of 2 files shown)


SoftLayer/CLI/hardware/credentials.py
```diff
@@ -22,6 +23,12 @@ def cli(env, identifier):
     instance = manager.get_hardware(hardware_id)
 
     table = formatting.Table(['username', 'password'])
+    if 'passwords' not in instance['operatingSystem']:
+        raise exceptions.SoftLayerError("No passwords found in operatingSystem")
+
     for item in instance['operatingSystem']['passwords']:
-        table.add_row([item['username'], item['password']])
```
allmightyspiff: instead of raising an exception, I would prefer this to just print out  "None"

    table.add_row([item.get('username', 'None'), item.get('password', 'None')])

.get() allows you to specify a default value to return if the item isn't found. This article explain it in a bit more detail if you are interested. 
https://docs.quantifiedcode.com/python-anti-patterns/correctness/not_using_get_to_return_a_default_value_from_a_dictionary.html

FernandoOjeda: Done

PR #993, opened 2018-07-05
remove deprecated arguments to prompt()

ScottieD:
prompt-toolkit has been updated many times since slcli-shell was created. p_shortcuts.prompt() passed arguments that are long deprecated causing the shell to fail.


ScottieD:
I removed the whole function and it passes all tests.

PR #995, opened 2018-07-05
New Feature to create a vm in a specific subnet

FernandoOjeda:
New Feature to create a vm in a specific subnet: https://github.com/softlayer/softlayer-python/issues/921

### DIFF
--- a/SoftLayer/CLI/virt/create.py
+++ b/SoftLayer/CLI/virt/create.py
@@ -121,7 +121,10 @@ def _parse_create_args(client, args):
     # Get the SSH keys
     if args.get('key'):
         keys = []
-        _add_keys(args, client, keys)
+        for key in args.get('key'):
+            resolver = SoftLayer.SshKeyManager(client).resolve_ids
+            key_id = helpers.resolve_id(resolver, key, 'SshKey')
+            keys.append(key_id)
         data['ssh_keys'] = keys
 
     if args.get('vlan_public'):
@@ -151,13 +154,6 @@ def _parse_create_args(client, args):
         data['host_id'] = args['host_id']
 
     return data
-
-
-def _add_keys(args, client, keys):
-    for key in args.get('key'):
-        resolver = SoftLayer.SshKeyManager(client).resolve_ids
-        key_id = helpers.resolve_id(resolver, key, 'SshKey')
-        keys.append(key_id)
 
 
 @click.command(epilog="See 'slcli vs create-options' for valid options")
--- a/SoftLayer/CLI/virt/create.py
+++ b/SoftLayer/CLI/virt/create.py
@@ -65,6 +65,7 @@ def _update_with_like_args(ctx, _, value):
         ctx.default_map = {}
     ctx.default_map.update(like_args)
 
+
 def _parse_create_args(client, args):
     """Converts CLI arguments to args for VSManager.create_instance.
 
--- a/SoftLayer/CLI/virt/create.py
+++ b/SoftLayer/CLI/virt/create.py
@@ -133,6 +133,12 @@ def _parse_create_args(client, args):
     if args.get('vlan_private'):
         data['private_vlan'] = args['vlan_private']
 
+    if args.get('subnet_public'):
+        data['public_subnet'] = args['subnet_public']
+
+    if args.get('subnet_private'):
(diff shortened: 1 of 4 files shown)


SoftLayer/CLI/virt/create.py
```diff
@@ -121,10 +121,7 @@ def _parse_create_args(client, args):
     # Get the SSH keys
     if args.get('key'):
         keys = []
-        for key in args.get('key'):
```
allmightyspiff: Was there any reason for moving these lines to their own function?

SoftLayer/CLI/virt/create.py
```diff
@@ -150,6 +153,13 @@ def _parse_create_args(client, args):
     return data
 
 
+def _add_keys(args, client, keys):
```
allmightyspiff: I don't like functions that modify the data that gets passed in.  this function should `return keys` instead of modifying the array that gets passed in.

SoftLayer/managers/vs.py
```diff
@@ -415,6 +413,46 @@ def _generate_create_dict(
 
         return data
 
+    def _create_network_components(
```
allmightyspiff: Try something like this instead.


```
parameters = {}
if private_vlan:
    parameters['primaryBackendNetworkComponent'] = { "networkVlan": {"id": int(private_vlan)}}
if public_vlan:
   parameters['primaryNetworkComponent'] =  { "networkVlan": {"id": int(public_vlan)}}
if public_subnet:
   if public_vlan is None:
       throw Excetions("You need to specify a public_vlan with public_subnet") 
   parameters['primaryNetworkComponent']['networkVlan]['primarySubnet'] = {'id' : int(public_subnet)}
if private_subnet:
   if private_vlan is None:
       throw Excetions("You need to specify a private_vlan with private_subnet") 
   parameters['primaryBackendNetworkComponent']['networkVlan]['primarySubnet'] = {'id' : int(private_subnet)}
```

SoftLayer/CLI/virt/create.py
FernandoOjeda: The problem was that it exceeded the maximum number of branches in the method and I only made a refactor of the for cycle in another method using the IDE's refactor options.

SoftLayer/CLI/virt/create.py
```diff
@@ -133,11 +133,9 @@ def _parse_create_args(client, args):
     if args.get('vlan_private'):
         data['private_vlan'] = args['vlan_private']
 
-    if args.get('subnet_public'):
-        data['public_subnet'] = args['subnet_public']
+    data['public_subnet'] = args.get('subnet_public', None)
```
FernandoOjeda: I tried to implement the switcher option for this method, but we get the ¨args¨ parameter, it's not a specific data. I was implementing like this:

    switcher = {
        args.get('san'): attribute_san(args, data),
        args.get('os'): attribute_os(args, data),
        args.get('image'): attribute_image(args, client, data),
        args.get('datacenter'): attribute_datacenter(args, data),
        args.get('network'): attribute_network(args, data),
        args.get('postinstall'): attribute_postinstall(args, data),
        args.get('key'): attribute_key(args, client, data),
        args.get('vlan_public'): attribute_vlan_public(args, data),
        args.get('vlan_private'): attribute_vlan_private(args, data),
        args.get('subnet_public'): attribute_subnet_public(args, data),
        args.get('subnet_private'): attribute_subnet_private(args, data),
        args.get('public_security_group'): attribute_public_security_group(args, data),
        args.get('private_security_group'): attribute_private_security_group(args, data),
        args.get('tag'): attribute_tag(args, data),
        args.get('host_id'): attribute_host_id(args, data),
    }

    return data


def attribute_host_id(args, data):
    if args.get('host_id'):
        data['host_id'] = args['host_id']


def attribute_tag(args, data):
    if args.get('tag'):
        data['tags'] = ','.join(args['tag'])


def attribute_private_security_group(args, data):
    if args.get('private_security_group'):
        priv_groups = args.get('private_security_group')
        data['private_security_groups'] = [group for group in priv_groups]
..........................

In this case the attribute switches will not be use, because in each method each data from args will be add to the data attribute.
That's why I used the first option you suggested.

FernandoOjeda:
Yes, it is working well.

Issue #1008, opened 2018-07-23
Ability to delete a user through SLCLI


fmiquiza:
**Please triple-check to make sure that you have properly masked out user credentials like usernames, passwords and API keys before submitting your issue**

### Expected Behavior
The ability to delete a user through the SLCLI

### Actual Behavior
Customers require to delete a user through the python Slcli 

### Environment Information
Operating System: 
softlayer-python version (`slcli --version`):

allmightyspiff:
```json
$ slcli --format=json call-api SoftLayer_User_Customer_Status getAllObjects
[
    { "id": 1001, "keyName": "ACTIVE", "name": "Active" },
    { "id": 1002, "keyName": "DISABLED", "name": "Disabled" },
    { "id": 1003, "keyName": "INACTIVE",  "name": "Inactive" },
    { "id": 1021, "keyName": "CANCEL_PENDING",  "name": "cancel_pending"  },
    { "id": 1022, "keyName": "VPN_ONLY",  "name": "VPN Only" }
]
```

Here is how to disable a user. I think statusID 1021 will eventually get the user deleted, but I'm not 100% sure about that. 

```shell
$ slcli user edit-details 282220 -t '{"userStatusId" :  1002}'
282220 updated successfully
$ slcli user detail  282220
:..............:.............................................:
:        Title : Basic Information                           :
:..............:.............................................:
:           Id : 282220                                      :
:     Username : test-miguel3                                :
:         Name : test test                                   :
:        Email : test@softlayer.com                          :
:       OpenID : None                                        :
:      Address : 315 Capitol Street None Houston TX US 77002 :
:      Company : SoftLayer Internal - Development Community  :
:      Created : 2014-11-05T12:17:13-06:00                   :
: Phone Number : None                                        :
:  Parent User : someone                                     :
:       Status : Disabled                                    :
:     PPTP VPN : False                                       :
:      SSL VPN : False                                       :
:..............:.............................................:
```

I'll leave this issue open since having a command to do this would be ideal.

allmightyspiff:
merged in, I'll release v5.5.1 friday with this included.

PR #1014, opened 2018-08-01
Adding user delete command and unittests

acamacho82:
The new command allows to delete a user through its identifier (issue #1008 )


SoftLayer/CLI/user/delete.py
```diff
@@ -0,0 +1,30 @@
...
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import helpers
+
+
+@click.command()
+@click.argument('identifier')
+@environment.pass_env
+def cli(env, identifier):
+    """Delete a User.
```
allmightyspiff: To be clear to the customer, I would change the help message to this:

"Sets a user's status to CANCEL_PENDING, which will immediately disable the account, and will eventually be fully removed from the account by an automated internal process"

acamacho82: updated, tox analysis should pass too

Issue #1006, opened 2018-07-17
Paginate list views


allmightyspiff:
- vlan list
- hw list
- vs list

Need a way to easily paginate these for when their results are > 100

PR #1015, opened 2018-08-01
Pagination for vs list and others

allmightyspiff:
#1006 

Added pagination to the following slcli command
- `vs list`
- `hw list`
- `vlan list`

Fixed iter_call so other commands can take advantage of this feature in the future.


acamacho82:
the changes looks good, I think this can be merged

Issue #1022, opened 2018-08-21
Unable to order a dedicatedhost from SLCLI for a GPU Flavor


FernandoOjeda:
Unable to order a dedicatedhost from slcli for a GPU flavor.

slcli dedicatedhost create -H E2ETestHost1 -D e2eperftest.com -f 56_CORES_X_484_RAM_X_1_5_TB_X_2_GPU_P100 -d dal13

Receive the below error when I submit an order request.

SoftLayerAPIError(SoftLayer_Exception_Public): Unable to find a price for the dedicated host configuration provided.

The slcli to get the routers available for the dedicatedHost item use the following structure:

SoftLayer_Virtual_DedicatedHost::getAvailableRouters

```
{
      "cpuCount": 56,
      "diskCapacity": 1200,
      "memoryCapacity": 360,
      "datacenter": {
      	"id": 1854895
      }

```
But for dedicatedHost GPU flavor has to add the following structure:

```
{
      "cpuCount": 56,
      "diskCapacity": 1200,
      "memoryCapacity": 360,
      "datacenter": {
      	"id": 1854895
      },
      "pciDevices": [ 
      	{ 
      		"hardwareComponentModel": 
      		   { 
      		   	   "hardwareGenericComponentModel":
			               {"hardwareComponentType":
					{"keyName":"GPU"
					}
				}
      		   	
      		   } 
      	}, 
      	{ 
      		"hardwareComponentModel": 
      		   { 
      		   	   "hardwareGenericComponentModel":
			               {"hardwareComponentType":
					{"keyName":"GPU"
					}
				}
        	 } 
       	} 
      ] 
    }

```
Environment Information
Operating System: 
softlayer-python version (`slcli --version`): version 5.5.0

kwienken:
@FernandoOjeda I don't think using the hardwareComponentType alone is enough since not all dedicated hosts use the same GPU model. You will need to get the model from the bundled GPU product. The bundled product also specifies the number of GPUs.

allmightyspiff:
@kwienken given just the flavor 56_CORES_X_484_RAM_X_1_5_TB_X_2_GPU_P100 how would we get the needed structure to call SoftLayer_Virtual_DedicatedHost::getAvailableRouters?

kwienken:
You'll need to get the product by keyName and use the mask ```
mask[
    capacity,
    units,
    hardwareGenericComponentId,
    itemCategory[categoryCode],
    bundleItems[
        capacity,
        units,
[...]

PR #1025, opened 2018-08-23
create dedicated host with gpu

FernandoOjeda:
Create dedicated host with gpu https://github.com/softlayer/softlayer-python/issues/1022.

### DIFF
--- a/SoftLayer/managers/dedicated_host.py
+++ b/SoftLayer/managers/dedicated_host.py
@@ -243,7 +243,8 @@ class DedicatedHostManager(utils.IdentifierMixin, object):
             capacity,
             keyName,
             itemCategory[categoryCode],
-            bundleItems[capacity, categories[categoryCode]]
+            bundleItems[capacity, keyName, categories[categoryCode], hardwareGenericComponentModel[id,
+            hardwareComponentType[keyName]]]
         ],
         regions[location[location[priceGroups]]]
         '''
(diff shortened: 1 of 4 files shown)


SoftLayer/managers/dedicated_host.py
```diff
@@ -317,18 +318,57 @@ def _get_backend_router(self, locations, item):
                 if category['categoryCode'] == 'dedicated_host_disk':
                     disk_capacity = capacity['capacity']
 
+        for hardwareComponent in item['bundleItems']:
+            if hardwareComponent['keyName'].find("GPU") != -1:
```
allmightyspiff: instead of creating 4 variables, it would makes sense to do something like this:

```python
gpuComponent =  {
                                    'hardwareComponentModel': {
                                        'hardwareGenericComponentModel': {
                                            'id': hardwareGenericComponentModelId,
                                            'hardwareComponentType': {
                                                'keyName': hardwareComponentTypeKeyName
                                            }
                                        }
                                    }
                                },
                                {
                                    'hardwareComponentModel': {
                                        'hardwareGenericComponentModel': {
                                            'id': hardwareGenericComponentModelId,
                                            'hardwareComponentType': {
                                                'keyName': hardwareComponentTypeKeyName
                                            }
                                        }
                                    }
                                }

```

Then just use `gpuComponent` further down.

SoftLayer/managers/dedicated_host.py
```diff
@@ -317,18 +318,57 @@ def _get_backend_router(self, locations, item):
...
                 if location['locationId'] is not None:
                     loc_id = location['locationId']
-                    host = {
-                        'cpuCount': cpu_count,
-                        'memoryCapacity': mem_capacity,
-                        'diskCapacity': disk_capacity,
-                        'datacenter': {
-                            'id': loc_id
+                    if item['keyName'].find("GPU") == -1:
+                        host = {
```
allmightyspiff: this bit should be outside of the `if` statement since it always need to be set, and is the same for GPU/NO-GPU

SoftLayer/managers/dedicated_host.py
```diff
@@ -317,18 +318,57 @@ def _get_backend_router(self, locations, item):
...
+                            'cpuCount': cpu_count,
+                            'memoryCapacity': mem_capacity,
+                            'diskCapacity': disk_capacity,
+                            'datacenter': {
+                                'id': loc_id
+                            }
+                        }
+                    else:
```
allmightyspiff: This `if/else` should instead just be a single `if gpuComponent is not None:` and then add the gpuComponent to the already existing `host` dictionary

SoftLayer/managers/dedicated_host.py
FernandoOjeda: Done

SoftLayer/managers/dedicated_host.py
FernandoOjeda: Done.

SoftLayer/managers/dedicated_host.py
FernandoOjeda: Done.

Issue #1033, opened 2018-09-04
Price estimate in slcli vs create --test omits the base price of the flavor


jtapsl:
### Expected Behavior
The price estimate when adding the --test flag to slcli vs create should show the correct retail price.

### Actual Behavior
The price estimate includes the price of add-ons such as disks, network speed, etc, but omits the base cost of the flavor.

$ slcli vs create --hostname pricetest --domain example.org --datacenter wdc01 --os CENTOS_7_64 --billing monthly --flavor B1_2X8X25 --private --test | tee /tmp/pricetest.txt
CentOS 7.x - Minimal Install (64 bit)                  0.00  
25 GB (SAN)                                            0.00  
Reboot / Remote Console                                0.00  
10 Mbps Private Network Uplink                         0.00  
0 GB Bandwidth Allotment                               0.00  
1 IP Address                                           0.00  
Host Ping and TCP Service Monitoring                   0.00  
Email and Ticket                                       0.00  
Automated Reboot from Monitoring                       0.00  
Unlimited SSL VPN Users & 1 PPTP VPN User per account  0.00  
Nessus Vulnerability Assessment & Reporting            0.00  
Total monthly cost                                     0.00  
 -- ! Prices reflected here are retail and do not take account level discounts and are not guaranteed.

### Environment Information
Operating System: Red Hat Enterprise Linux Workstation release 7.5 (Maipo)
softlayer-python version (`slcli --version`): slcli (SoftLayer Command-line), version 5.5.2

allmightyspiff:
This fix will be included in the 5.5.3 release

PR #1036, opened 2018-09-07
vs create flavor test

FernandoOjeda:
Fixed vs create with flavor data test https://github.com/softlayer/softlayer-python/issues/1033

### DIFF
--- a/SoftLayer/CLI/virt/create.py
+++ b/SoftLayer/CLI/virt/create.py
@@ -273,11 +273,13 @@ def cli(env, **args):
 
         if result['presetId']:
             ordering_mgr = SoftLayer.OrderingManager(env.client)
+            item_prices = ordering_mgr.get_item_prices(result['packageId'])
             preset_prices = ordering_mgr.get_preset_prices(result['presetId'])
             search_keys = ["guest_core", "ram"]
             for price in preset_prices['prices']:
                 if price['item']['itemCategory']['categoryCode'] in search_keys:
-                    result['prices'].append(price)
+                    item_key_name = price['item']['keyName']
+                    _add_item_prices(item_key_name, item_prices, result)
 
         table = _build_receipt_table(result['prices'], args.get('billing'))
 
@@ -316,6 +318,16 @@ def cli(env, **args):
                 raise exceptions.CLIHalt(code=1)
 
     env.fout(output)
+
+
+def _add_item_prices(item_key_name, item_prices, result):
+    """Add the flavor item prices to the rest o the items prices"""
+    for item in item_prices:
+        if item_key_name == item['item']['keyName']:
+            if 'pricingLocationGroup' in item:
+                for location in item['pricingLocationGroup']['locations']:
+                    if result['location'] == str(location['id']):
+                        result['prices'].append(item)
 
 
 def _build_receipt_table(prices, billing="hourly"):
--- a/SoftLayer/CLI/virt/create.py
+++ b/SoftLayer/CLI/virt/create.py
@@ -282,11 +282,11 @@ def cli(env, **args):
         if str(result['presetId']) is not "":
             ordering_mgr = SoftLayer.OrderingManager(env.client)
             preset_prices = ordering_mgr.get_preset_prices(result['presetId'])
-            rate, total_preset_hourly, total_preset_monthly = get_total_recurring_fee(args, preset_prices, table,
-                                                                                      total_preset_hourly,
-                                                                                      total_preset_monthly)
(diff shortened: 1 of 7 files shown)


SoftLayer/CLI/virt/create.py
```diff
@@ -334,6 +335,20 @@ def cli(env, **args):
     env.fout(output)
 
 
+def get_total_recurring_fee(args, result, table, total_hourly, total_monthly):
```
allmightyspiff: I don't like functions that modify the arguments that get passed in, good way to introduce bugs in my opinion. 

Lets change this function to `_build_reciept_table(prices)` which will build and return the whole table used to print costs. It doesn't need to know about hourly v monthly.

```python
def get_total_recurring_fee(prices, billing="hourly"):
    """Retrieve the total recurring fee of the items prices"""
    total = 0.0
    table = formatting.Table(['Cost', 'Item'])
    table.align['Cost'] = 'r'
    table.align['Item'] = 'l'
    for price in prices:
        rate = 0.0
        if billing == "hourly":
            rate += float(price.get('hourlyRecurringFee', 0.0))
        else:
            rate += float(price.get('recurringFee', 0.0))
        total += rate

        table.add_row(["%.3f" % rate, price['item']['description']])
    table.add_row(["%.3f" % total, "Total %s cost" % billing])
    return table
```

**prices need to be to 3 decimals**

SoftLayer/CLI/virt/create.py
```diff
@@ -272,26 +272,27 @@ def cli(env, **args):
...
-        for price in result['prices']:
-            total_monthly += float(price.get('recurringFee', 0.0))
-            total_hourly += float(price.get('hourlyRecurringFee', 0.0))
-            if args.get('billing') == 'hourly':
-                rate = "%.2f" % float(price['hourlyRecurringFee'])
-            elif args.get('billing') == 'monthly':
-                rate = "%.2f" % float(price['recurringFee'])
+        if result['presetId']:
+            ordering_mgr = SoftLayer.OrderingManager(env.client)
+            preset_prices = ordering_mgr.get_preset_prices(result['presetId'])
```
allmightyspiff: If they are ordering a preset, what you should do is get the preset prices, and then add the server and ram items to the `results['prices']` array. It looks like the receipt already has the disk prices so no worries there.  

```python
if result['presetId']:
    ordering_mgr = SoftLayer.OrderingManager(env.client)
    preset_prices = ordering_mgr.get_preset_prices(result['presetId'])
    search_keys = ["guest_core", "ram"]
    for price in preset_prices['prices']:
        if price['item']['itemCategory']['categoryCode'] in search_keys:
            result['prices'].append(price)
```

allmightyspiff: Why the result data doesn't have the core/ram info is a mystery to me, but that was the best I could think of to get that data but NOT get all the other price data that gets returned.

allmightyspiff:
While testing I made some changes that I think work, your welcome to apply them if it makes it easier
[patch.txt](https://github.com/softlayer/softlayer-python/files/2368672/patch.txt)

allmightyspiff:
screen shot 2018-09-10 at 6 01 32 pm


The price for RAM+CPU from the API doens't seem to match up with the portal.
Portal shows price for BL2_4X16X300 though.... I'll have to see whats going on with that.

tests/managers/ordering_tests.py
```diff
@@ -68,6 +68,57 @@ def test_get_package_id_by_type_returns_valid_id(self):
 
         self.assertEqual(46, package_id)
 
+    def test_get_preset_prices(self):
+        preset_id = 405
+        preset_prices = self.ordering.get_preset_prices(preset_id)
+
+        self.assertEqual(preset_prices, {
```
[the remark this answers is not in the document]
FernandoOjeda: Done.

SoftLayer/CLI/virt/create.py
```diff
@@ -272,26 +272,27 @@ def cli(env, **args):
         result = vsi.verify_create_instance(**data)
         total_monthly = 0.0
         total_hourly = 0.0
+        total_preset_monthly = 0.0
+        total_preset_hourly = 0.0
```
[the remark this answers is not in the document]
FernandoOjeda: Done.

SoftLayer/CLI/virt/create.py
FernandoOjeda: Done.

SoftLayer/CLI/virt/create.py
FernandoOjeda: Done.

FernandoOjeda:
Now the total cost is right.
image

Issue #584, opened 2015-07-20
Cannot create a reverse pointer DNS record with slcli


TyMac:
The following command returns no error but no ptr record seems to be created:

slcli dns record-add domain001.com 233 ptr server0001.damain001.com.

233 is the last octet in the IP address. The ongoing thread in the softlayer forums is located here:

https://forums.softlayer.com/forum/softlayer-developer-network/general-discussion/84322-error-when-attempting-to-use-slcli-to-create-a-reverse-pointer-dns-entry

sudorandom:
Interesting... so PTR records don't come back from the API the same way that other records do... That seems weird.

This is an API call that can be used to see PTR records: http://sldn.softlayer.com/search/apachesolr_search/getReverseDomainRecords

Just a note, there is a command that enabled you to sync dns records (for virtual servers only). Example as follows:

```bash
$ slcli vs dns-sync some-host
```

The source code for that lives here: https://github.com/softlayer/softlayer-python/blob/3eafa595e7615200a2e5a3b0dd0bdfebcbe7f499/SoftLayer/CLI/virt/dns.py#L60-L80

There still is an issue that PTR records don't seem very visible from the CLI at this time.

briancline:
There was a time when I spent about 4 hours looking into the API calls for PTR records for both IPv4 and IPv6 zones. Creating a PTR for an IPv6 zone through our API was quite an endurance test, but got it to work through the Python client eventually.

Might be hairy to create a user-friendly way to specify a zone without having to give the full in-addr.arpa or ip6.arpa zone names, but could probably allow a CIDR to help interpret the proper zone name. Seems necessary for specifying ip6.arpa zone names though because it gets really tedious and error-prone really quickly.

Either way, in this case, it sounds like the desire is to allow the existing commands handle PTRs as expected in a normal zone (a typical case when someone's used created CNAME in a .in-addr.arpa zone to delegate an individual PTR to you, by using a CNAME to point to some predetermined record in your zone).

I'll give this a shot to get it added to the DNS manager and the CLI.

dominikkukacka:
After the change will the feature also be available in the SDK?

PR #1037, opened 2018-09-07
New methods in DNS manager

acamacho82:
The new methods allows to create MX, SRV and PTR records.
The CLI supports the creation of A, AAAA, MX, TXT, SPF, SRV, CNAME, and PTR records, it should fix the issue #584

### DIFF
--- a/SoftLayer/CLI/dns/record_add.py
+++ b/SoftLayer/CLI/dns/record_add.py
@@ -75,19 +75,16 @@ def cli(env, record, type, data, zone, ttl, priority, protocol, port, service, w
         zone_id = helpers.resolve_id(manager.resolve_ids, zone, name='zone')
 
         if type == 'MX':
-            result = manager.create_record_mx(zone_id, record, data, ttl=ttl, priority=priority)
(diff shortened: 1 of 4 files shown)


SoftLayer/CLI/dns/record_add.py
```diff
@@ -5,24 +5,86 @@
...
+        slcli dns record-add myhost.com SPF "v=spf1 include:_spf.google.com ~all" --zone=2223334
+
+        slcli dns record-add myhost.com SRV 192.168.1.10 --zone=2223334 --service=foobar --port=80 --protocol=TCP
+
+    """
 
     manager = SoftLayer.DNSManager(env.client)
-    zone_id = helpers.resolve_id(manager.resolve_ids, zone, name='zone')
-    manager.create_record(zone_id, record, type, data, ttl=ttl)
+    type = type.upper()
```
allmightyspiff: [`type`](https://docs.python.org/3/library/functions.html#type) is a built in function of python, use `zone_type` or something similar instead. 
I know it was already `type`, but now is a good time to fix that I think.

acamacho82:
Agree, I changed type by record_type :D

PR #1053, opened 2018-10-05
Icos image support

mikewurtz:
Add export/import capabilities to/from IBM Cloud Object Storage to the image manager as well as the slcli. Unit tests updated to include tests for COS path.

### DIFF
--- a/SoftLayer/CLI/image/export.py
+++ b/SoftLayer/CLI/image/export.py
@@ -12,7 +12,7 @@ from SoftLayer.CLI import helpers
 @click.command()
 @click.argument('identifier')
 @click.argument('uri')
-@click.option('--ibm_api_key',
+@click.option('--ibm-api-key',
               default="",
               help="The IBM Cloud API Key with access to IBM Cloud Object "
                    "Storage instance.")
--- a/SoftLayer/CLI/image/export.py
+++ b/SoftLayer/CLI/image/export.py
@@ -15,7 +15,9 @@ from SoftLayer.CLI import helpers
 @click.option('--ibm-api-key',
               default="",
               help="The IBM Cloud API Key with access to IBM Cloud Object "
-                   "Storage instance.")
+                   "Storage instance. For help creating this key see "
+                   "https://console.bluemix.net/docs/services/cloud-object-"
+                   "storage/iam/users-serviceids.html#serviceidapikeys")
 @environment.pass_env
 def cli(env, identifier, uri, ibm_api_key):
     """Export an image to object storage.
--- a/SoftLayer/CLI/image/export.py
+++ b/SoftLayer/CLI/image/export.py
@@ -12,17 +12,23 @@ from SoftLayer.CLI import helpers
 @click.command()
 @click.argument('identifier')
 @click.argument('uri')
+@click.option('--ibm_api_key',
+              default="",
+              help="The IBM Cloud API Key with access to IBM Cloud Object "
+                   "Storage instance.")
 @environment.pass_env
-def cli(env, identifier, uri):
+def cli(env, identifier, uri, ibm_api_key):
     """Export an image to object storage.
 
     The URI for an object storage object (.vhd/.iso file) of the format:
     swift://<objectStorageAccount>@<cluster>/<container>/<objectPath>
+    or cos://<clusterName>/<bucketName>/<objectPath> if using IBM Cloud
+    Object Storage
     """
 
     image_mgr = SoftLayer.ImageManager(env.client)
     image_id = helpers.resolve_id(image_mgr.resolve_ids, identifier, 'image')
-    result = image_mgr.export_image_to_uri(image_id, uri)
+    result = image_mgr.export_image_to_uri(image_id, uri, ibm_api_key)
 
     if not result:
         raise exceptions.CLIAbort("Failed to export Image")
--- a/SoftLayer/CLI/image/export.py
+++ b/SoftLayer/CLI/image/export.py
@@ -13,7 +13,7 @@ from SoftLayer.CLI import helpers
 @click.argument('identifier')
 @click.argument('uri')
 @click.option('--ibm-api-key',
-              default="",
+              default=None,
               help="The IBM Cloud API Key with access to IBM Cloud Object "
                    "Storage instance. For help creating this key see "
                    "https://console.bluemix.net/docs/services/cloud-object-"
--- a/SoftLayer/CLI/image/import.py
+++ b/SoftLayer/CLI/image/import.py
@@ -18,13 +18,38 @@ from SoftLayer.CLI import formatting
 @click.option('--os-code',
               default="",
               help="The referenceCode of the operating system software"
-                   " description for the imported VHD")
+                   " description for the imported VHD, ISO, or RAW image")
+@click.option('--ibm-api-key',
+              default="",
+              help="The IBM Cloud API Key with access to IBM Cloud Object "
+                   "Storage instance.")
+@click.option('--root-key-id',
+              default="",
+              help="ID of the root key in Key Protect")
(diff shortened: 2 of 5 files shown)


SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
+@click.option('--root-key-id',
+              default="",
+              help="ID of the root key in Key Protect")
+@click.option('--wrapped-dek',
+              default="",
+              help="Wrapped Decryption Key provided by IBM KeyProtect")
+@click.option('--kp-id',
+              default="",
+              help="ID of the IBM Key Protect Instance")
+@click.option('--cloud-init',
```
allmightyspiff: Click has an option `is_flag=True,` which should be used here. This way you don't need the `default=""` bit

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
 @click.option('--os-code',
               default="",
               help="The referenceCode of the operating system software"
-                   " description for the imported VHD")
+                   " description for the imported VHD, ISO, or RAW image")
+@click.option('--ibm-api-key',
```
allmightyspiff: I'm not terribly familiar with IBM COS, but are these values obvious to someone who has used IBM COS? If not could you add some sort of info on where to get the various keys?

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
+              default="",
+              help="The IBM Cloud API Key with access to IBM Cloud Object "
+                   "Storage instance.")
+@click.option('--root-key-id',
+              default="",
+              help="ID of the root key in Key Protect")
+@click.option('--wrapped-dek',
+              default="",
+              help="Wrapped Decryption Key provided by IBM KeyProtect")
```
fode-ibm: Change to, "Wrapped Data Encryption Key provided by IBM KeyProtect".

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
+@click.option('--kp-id',
+              default="",
+              help="ID of the IBM Key Protect Instance")
+@click.option('--cloud-init',
+              default="",
```
fode-ibm: default value should be Boolean, which should change this option type to Boolean as well

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
+              help="Specifies if image is cloud init")
+@click.option('--byol',
+              default="",
```
fode-ibm: default value should be Boolean, which should change this option type to Boolean as well.

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
+              help="Specifies if image is bring your own license")
+@click.option('--is-encrypted',
+              default="",
```
fode-ibm: default value should be Boolean

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
+              help="ID of the root key in Key Protect")
+@click.option('--wrapped-dek',
+              default="",
+              help="Wrapped Decryption Key provided by IBM KeyProtect")
+@click.option('--kp-id',
+              default="",
+              help="ID of the IBM Key Protect Instance")
+@click.option('--cloud-init',
+              default="",
+              help="Specifies if image is cloud init")
```
fode-ibm: The term, "cloud-init" should have hyphen, as it is the name of a specification. Change to, "Specifies whether image is enabled for cloud initialization using cloud-init".

SoftLayer/managers/image.py
```diff
@@ -120,28 +120,68 @@ def edit(self, image_id, name=None, note=None, tag=None):
 
         return bool(name or note or tag)
 
-    def import_image_from_uri(self, name, uri, os_code=None, note=None):
+    def import_image_from_uri(self, name, uri, os_code=None, note=None,
+                              ibm_api_key=None, root_key_id=None,
+                              wrapped_dek=None, kp_id=None, cloud_init=None,
+                              byol=None, is_encrypted=None):
```
fode-ibm: Default values for these cloud_init, byol, and is_encrypted params should be Boolean False values, which would then match the description for them below.

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
 @click.option('--os-code',
               default="",
```
fode-ibm: Consider changing default to None, rather than empty string.

SoftLayer/CLI/image/export.py
```diff
@@ -12,17 +12,23 @@
 @click.command()
 @click.argument('identifier')
 @click.argument('uri')
+@click.option('--ibm-api-key',
+              default="",
```
fode-ibm: Consider changing default to None, rather than empty string.

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
               help="The referenceCode of the operating system software"
-                   " description for the imported VHD")
+                   " description for the imported VHD, ISO, or RAW image")
+@click.option('--ibm-api-key',
+              default="",
```
fode-ibm: Consider changing default to None, rather than empty string.

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
+              help="The IBM Cloud API Key with access to IBM Cloud Object "
+                   "Storage instance.")
+@click.option('--root-key-id',
+              default="",
```
fode-ibm: Consider changing default to None, rather than empty string.

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
+              help="ID of the root key in Key Protect")
+@click.option('--wrapped-dek',
+              default="",
+              help="Wrapped Decryption Key provided by IBM KeyProtect")
+@click.option('--kp-id',
+              default="",
```
fode-ibm: Consider changing default to None, rather than empty string.

SoftLayer/CLI/image/import.py
```diff
@@ -18,13 +18,38 @@
...
+                   " description for the imported VHD, ISO, or RAW image")
+@click.option('--ibm-api-key',
+              default="",
+              help="The IBM Cloud API Key with access to IBM Cloud Object "
+                   "Storage instance.")
+@click.option('--root-key-id',
+              default="",
+              help="ID of the root key in Key Protect")
+@click.option('--wrapped-dek',
+              default="",
```
fode-ibm: Consider changing default to None, rather than empty string.

SoftLayer/CLI/image/import.py
mikewurtz: The default values for these flags are going to be removed in the next patch. `is_flag=True,` removes the need to have a default value.

SoftLayer/managers/image.py
mikewurtz: This will be included in the next patch.

SoftLayer/CLI/image/import.py
mikewurtz: Will change in the next commit.

SoftLayer/CLI/image/import.py
mikewurtz: I will add some links to our documentation on the ibm-api-key and the wrapped-dek.

SoftLayer/CLI/image/export.py
mikewurtz: Will add on next commit.

PR #1071, opened 2018-11-07
NETWORK-8987 - added createDate and modifyDate parameters to sg rule-list

Anjana-Rajagopal:
This adds the createDate and modifyDate parameters to the sg rule-list table under rule.py and to the mask under the get_securitygroup command under network.py.


Anjana-Rajagopal:
I modified the failing method in the securitygroup_test.py class to have the createDate and modifyDate parameters, as the error appeared to be a detected mismatch. My change has been added to this pull request.

Issue #1041, opened 2018-09-28
How to cancel dedicatehost in CLI?


yhuanghamu:
**Please triple-check to make sure that you have properly masked out user credentials like usernames, passwords and API keys before submitting your issue**
I know we can cancel it in web portal, but there is no `slcli dedicatedhost cancel` command option.
### Environment Information
Operating System: Mac OSX 10.13.6
softlayer-python version (`slcli --version`): 5.5.1

acamacho82:
Currently there is no way to delete dedicated hosts using the command `slcli dedicatedhost` but you should be able to do it through `slcli call-api`

```
$ slcli call-api Virtual_DedicatedHost deleteObject --id=10201
true
```
To retrieve the list of dedicated host and their ids use `slcli dedicatedhost list`

yhuanghamu:
Nice. Thanks.

yhuanghamu:
Cancellation of Dedicated hosts relied on the cancellation of attached virtual servers. 
Could you add a cascading option? Which means if i am going to cancel a dedicated host then all virtual servers running on top of it will be cancelled as well.
Another feature request is to make it async, which means i can open two tickets, first ticket to cancel virtual servers,  second ticket to cancel dedicated host. Second ticket will be processed after virtual servers are killed.

allmightyspiff:
Merged into master, this will be included in 5.6.4

PR #1075, opened 2018-11-16
dedicatedhost cancel, cancel-guests, and list-guests commands

acamacho82:
`slcli dh cancel` makes a simple call to the deleteObject method of dedicated host. 
`slcli dh cancel-guests` cancels all guests into the dedicated hosts
`slcli dh list-guests`  lists guests in the dedicated host, it should be possible to filter, sortby and display more columns in the table

It should cover the request #1041

### DIFF
--- a/SoftLayer/CLI/dedicatedhost/cancel_guests.py
+++ b/SoftLayer/CLI/dedicatedhost/cancel_guests.py
@@ -26,9 +26,18 @@ def cli(env, identifier):
     if not (env.skip_confirmations or formatting.no_going_back(host_id)):
         raise exceptions.CLIAbort('Aborted')
 
+    table = formatting.Table(['id', 'server name', 'status'])
+
     result = dh_mgr.cancel_guests(host_id)
 
-    if result is True:
-        click.secho('All guests into the dedicated host %s were cancelled' % host_id, fg='green')
+    if result:
+        for status in result:
+            table.add_row([
+                status['id'],
+                status['fqdn'],
+                status['status']
+            ])
+
+        env.fout(table)
     else:
         click.secho('There is not any guest into the dedicated host %s' % host_id, fg='red')
--- a/SoftLayer/CLI/dedicatedhost/cancel_guests.py
+++ b/SoftLayer/CLI/dedicatedhost/cancel_guests.py
@@ -1,0 +1,39 @@
+"""Cancel a dedicated host."""
+# :license: MIT, see LICENSE for more details.
+
+import click
+
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import exceptions
+from SoftLayer.CLI import formatting
+from SoftLayer.CLI import helpers
+
+
+@click.command()
+@click.argument('identifier')
+@environment.pass_env
+def cli(env, identifier):
+    """Cancel all virtual guests of the dedicated host immediately"""
+
+    dh_mgr = SoftLayer.DedicatedHostManager(env.client)
+    vs_mgr = SoftLayer.VSManager(env.client)
+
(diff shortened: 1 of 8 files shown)


acamacho82:
It failed due to my latest commit, I will fix it

SoftLayer/managers/dedicated_host.py
```diff
@@ -33,10 +33,138 @@ def __init__(self, client, ordering_manager=None):
...
+        :return: True on success or an exception
+
+        Example::
+            # Cancels dedicated host id 12345
+            result = mgr.cancel_host(12345)
+
+        """
+        return self.host.deleteObject(id=host_id)
+
+    def cancel_guests(self, host_id):
```
allmightyspiff: For this function, I'd like it to be able to return a structure that has a bit more information about what guests were canceled so the `slcli` can print out a receipt.

Ideally id, FQDN, and whether the guest was canceled or got an error

```
return  = [
    {'id': 1234, 'fqdn': 'test01.test.com', 'status': 'canceled' },
    {'id': 1235, 'fqdn': 'test02.test.com', 'status': 'Exception: Running Transaction or something' },
]

File: SoftLayer/managers/dedicated_host.py [hunk 2]
```diff
@@ -33,10 +33,138 @@ def __init__(self, client, ordering_manager=None):
...
+            # Cancel guests of dedicated host id 12345
+            result = mgr.cancel_guests(12345)
+        """
+        result = False
+
+        guests = self.host.getGuests(id=host_id, mask='id')
+
+        if guests:
+            for vs in guests:
+                result = self.guest.deleteObject(id=vs['id'])
```
allmightyspiff: Need to wrap this in a try/catch for SoftLayer.Exceptions. I can see a case where one guest will have a running transaction and won't be able to be canceled. So we don't want this command to fail midway.

SoftLayer/CLI/dedicatedhost/cancel_guests.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+
+    dh_mgr = SoftLayer.DedicatedHostManager(env.client)
+
+    host_id = helpers.resolve_id(dh_mgr.resolve_ids, identifier, 'dedicated host')
+
+    if not (env.skip_confirmations or formatting.no_going_back(host_id)):
+        raise exceptions.CLIAbort('Aborted')
+
+    result = dh_mgr.cancel_guests(host_id)
```
allmightyspiff: After updated the manager function, just add a way to print out a nice table with the results of each guests' cancelation.

SoftLayer/managers/dedicated_host.py
acamacho82: Done, it returns a dictionary and it is empty if there isn't any guest in the dedicated host. The `slcli` prints a table or a message in case the dict is empty.

SoftLayer/managers/dedicated_host.py
acamacho82: Done, is in the method called `_delete_guest`. I also added a unit-test to verify the exception message is retrieved in case the deleteObject fails

SoftLayer/CLI/dedicatedhost/cancel_guests.py
acamacho82: Done. It prints a table like below

```
$ slcli --really dh cancel-guests 33333
:..........:...........................:..................................................................:
:    id    :        server name        :                              status                              :
:..........:...........................:..................................................................:
: 1111111  :    foobar.softlayer.com   :                            Cancelled                             :
: 2222222  :    wombat.softlayer.com   : Exception: A billing item is required to process a cancellation. :
:..........:...........................:..................................................................:

```

Issue #1059, opened 2018-10-18
Support for SoftLayer_Hardware_Server::toggleManagementInterface


dnwe:
### Expected Behavior

The [SoftLayer_Hardware_Server::toggleManagementInterface](https://softlayer.github.io/reference/services/SoftLayer_Hardware_Server/toggleManagementInterface/) API call should be available in the softlayer-python client to provide feature parity with the [softlayer-go](https://github.com/softlayer/softlayer-go/issues/116) client.

### Actual Behavior

The [SoftLayer_Hardware_Server::toggleManagementInterface](https://softlayer.github.io/reference/services/SoftLayer_Hardware_Server/toggleManagementInterface/) is not currently supported.

### Environment Information
Operating System: all
softlayer-python version (`slcli --version`): 5.6.0

acamacho82:
I recommend try with the following code lines:

```python
hwId = 1122334 
client['Hardware_Server'].toggleManagementInterface(False, id=hwId)
```

You can also use the `slcli call-api` command to toggle the IPMI interface.

```bash
$ slcli call-api Hardware_Server toggleManagementInterface --id=1122334  True

```

PR #1080, opened 2018-11-19
Fix issue 1059

ATGE:


SoftLayer/CLI/hardware/toggle_ipmi.py
```diff
@@ -0,0 +1,23 @@
...
+import click
+
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import helpers
+
+
+@click.command()
+@click.argument('identifier')
+@click.option('--enabled',
```
allmightyspiff: You can use a [Boolean Flag](https://click.palletsprojects.com/en/7.x/options/#boolean-flags) here, so the user doesn't have to actually type in `enabled=True`.

See https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/CLI/user/edit_permissions.py#L13 for an example.

ATGE: I fixed it as following :
```
@click.option('--enable/--disable', default=True,
              help="Whether enable (DEFAULT) or disable the interface.")
```

allmightyspiff:
Resolves #1059

PR #1087, opened 2018-12-06
#676 - ipv6 support for creating virtual guests

### LINKED ISSUE #676: Cannot create virtual guest with ipv6 assigned
This is similar to  #410, but then for virtual guests.

allmightyspiff:
Refactored `slcli vs create` to use `SoftLayer_Product_Order::placeOrder` instead of `SoftLayer_Virtual_Guest::createObject` so that VSIs can be ordered with ipv6. 

Fixes #676


SoftLayer/CLI/virt/create.py
```diff
@@ -155,196 +143,110 @@ def _parse_create_args(client, args):
...
-@click.option('--template', '-t',
-              is_eager=True,
-              callback=template.TemplateCallback(list_args=['disk',
-                                                            'key',
-                                                            'tag']),
+@click.option('--template', '-t', is_eager=True,
+              callback=template.TemplateCallback(list_args=['disk', 'key', 'tag']),
               help="A template file that defaults the command-line options",
               type=click.Path(exists=True, readable=True, resolve_path=True))
 @click.option('--userdata', '-u', help="User defined metadata string")
```
acamacho82: It seems the `--userData` is broken, I sent the following call and the SoftLayer_Virtual_Guest::getUserData returns an empty value, it doesn't happens with the current version.

```slcli vs create --hostname ajcb01-testslxi --domain softlayer.com --datacenter dal09 --ipv6 --os UBUNTU_LATEST -f BL2_1X2X100 --billing hourly -i http://local.test/image.txt --disk={100,25} -n 1000 --private --userdata="name:asdkjh, ajksdhkjhk" --public-security-group={1064939,128317,1085497} --private-security-group={1085497,648345}```

File: SoftLayer/CLI/virt/create.py [hunk 2]
```diff
@@ -155,196 +143,110 @@ def _parse_create_args(client, args):
...
-    if args['export']:
-        export_file = args.pop('export')
-        template.export_to_template(export_file, args,
-                                    exclude=['wait', 'test'])
-        env.fout('Successfully exported options to a template file.')
-
-    if do_create:
-        if not (env.skip_confirmations or formatting.confirm(
-                "This action will incur charges on your account. Continue?")):
-            raise exceptions.CLIAbort('Aborting virtual server order.')
```
acamacho82: I couldn't see the prompt message: "This action will incur charges on your account. Continue?"

Is it not required now?

SoftLayer/CLI/virt/create.py
allmightyspiff: Thanks, I'll double check that. Likely just needs to be in a slightly different format now that im using placeOrder.

SoftLayer/CLI/virt/create.py
allmightyspiff: definitely didn't mean to remove this, thanks.

SoftLayer/CLI/virt/create.py
acamacho82: userData is working. I made a little more tests and it seems all is working so I think it can be merged.

PR #1099, opened 2019-02-05
Security groups request ids

erick-sapp:
This completes some of the requested work on the other current pull request.


allmightyspiff:
Finishes up #935

SoftLayer/CLI/event_log/get.py
```diff
@@ -0,0 +1,47 @@
...
+import click
+
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+
+COLUMNS = ['event', 'label', 'date', 'metadata']
+
+
+@click.command()
```
allmightyspiff: TODO:
Add a way to limit results. Something like in `slcli vs list` maybe.

SoftLayer/CLI/securitygroup/event_log.py
```diff
@@ -0,0 +1,32 @@
...
+
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+
+COLUMNS = ['event', 'label', 'date', 'metadata']
+
+
+@click.command()
+@click.argument('request_id')
```
allmightyspiff: TODO: I'm not really sure using the request_id here is useful, as that doesn't seem to be a very accessible piece of information for a user to get. Maybe it would be better to get the logs by the security group id?

SoftLayer/managers/event_log.py
```diff
@@ -0,0 +1,104 @@
...
+    """Provides an interface for the SoftLayer Event Log Service.
+
+    See product information here:
+    http://sldn.softlayer.com/reference/services/SoftLayer_Event_Log
+    """
+
+    def __init__(self, client):
+        self.event_log = client['Event_Log']
+
+    def get_event_logs(self, request_filter):
```
allmightyspiff: TODO: event_log.getAllObjects  should be called with `self.event_log.getAllObjects(filter=request_filter, iter=True)` and likely needs to support limits so more than the default number of objects can be pulled down.

SoftLayer/CLI/event_log/get.py
```diff
@@ -0,0 +1,47 @@
...
+
+import json
+
+import click
+
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+
+COLUMNS = ['event', 'label', 'date', 'metadata']
```
acamacho82: - It is a little annoying when looking the table and there are metadata information on it, maybe it would be better if metadata is an optional column or you could also create a new command which could be used to see the metadata information through the  `traceId` value, similar to `slcli vs detail`

- I suggest you to remove the braces {  } if metadata information will be in the table. 

- I suggest to rename `label` by `object` and recommend to add `object type` and `username` columns to the table, this would help the customers to identify what user made the action/event and in what kind of object, for example I remember that a customer wanted to know what user powered on/off an specific virtual server

SoftLayer/CLI/event_log/get.py
```diff
@@ -0,0 +1,47 @@
...
+              help="The type of the object we want to get event logs for")
+@click.option('--utc-offset', '-z',
+              help="UTC Offset for searching with dates. The default is -0000")
+@environment.pass_env
+def cli(env, date_min, date_max, obj_event, obj_id, obj_type, utc_offset):
+    """Get Event Logs"""
+    mgr = SoftLayer.EventLogManager(env.client)
+
+    request_filter = mgr.build_filter(date_min, date_max, obj_event, obj_id, obj_type, utc_offset)
+    logs = mgr.get_event_logs(request_filter)
```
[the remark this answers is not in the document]
erick-sapp: This should be taken care of with the latest push.

SoftLayer/CLI/event_log/get.py
erick-sapp: With commit d21cbd1, these suggestions should be implemented.

PR #1103, opened 2019-02-18
Event Log improvements

erick-sapp:
This adds limits and a couple fixes I found when playing with some other parts of the code.


SoftLayer/CLI/event_log/get.py
```diff
@@ -43,11 +46,19 @@ def cli(env, date_min, date_max, obj_event, obj_id, obj_type, utc_offset, metada
...
         user = log['userType']
+        label = ''
+
+        try:
+            label = log['label']
+        except KeyError:
+            pass  # label is already at default value.
+
         if user == "CUSTOMER":
             user = usrmgr.get_user(log['userId'], "mask[username]")['username']
```
allmightyspiff: Just noticed in testing that this line generates a large number of redundant API calls. Would be worth cacheing the results so we only have to look up each user once. I don't want to change this to just get all users once because some accounts have a lot of users which would generate its own problems. 

If you think thats out of the scope of this pull request thats ok, I can make an issue about it and deal with it later.

allmightyspiff:
I'm going to resolve this in #1125

PR #1106, opened 2019-02-27
Firmware reflash

follower46:
This allows users to reflash hardware firmware directly from the CLI.


SoftLayer/managers/hardware.py
```diff
@@ -613,6 +613,31 @@ def update_firmware(self,
         return self.hardware.createFirmwareUpdateTransaction(
             bool(ipmi), bool(raid_controller), bool(bios), bool(hard_drive), id=hardware_id)
 
+    def reflash_firmware(self,
+                        hardware_id,
```
allmightyspiff: should align with `self` not the `(`

Issue #1112, opened 2019-03-11
Can not upgrade an iops from 2 to 4 for a file storage endurance which has 24 TB volume


FernandoOjeda:
A customer wanted to upgrade an iops from 2 to 4 for a file storage endurance which has 24 TB volume:

The slcli command used was:
slcli file volume-modify --new-tier 4 11111

**Expected Behavior**
The request should upgrade the file storage endurance with the iops required.

**Actual Behavior**
The response of the request was:
Argument Error: Could not find price for endurance storage space

I tried the request by rest call and it worked.
**Environment Information**
Operating System: Windows 10 
softlayer-python version (`slcli --version`): 5.7.1

PR #1114, opened 2019-03-12
File storage endurance iops upgrade .

FernandoOjeda:
File storage endurance iops upgrade https://github.com/softlayer/softlayer-python/issues/1112.


allmightyspiff:
Fixes #1112

PR #1135, opened 2019-04-17
Use default retry logic in requests adapter (which is zero).

jsiembida:
The original commit doesn't explain why retry=3 was introduced in the first place.
Before, it was default, zero. Now it seems to conflict with retry decorator used
in some call in managers code resulting in 16 retries in total when combined.

### DIFF
--- a/SoftLayer/transports.py
+++ b/SoftLayer/transports.py
@@ -13,7 +13,6 @@ import time
 
 import requests
 from requests.adapters import HTTPAdapter
-from urllib3.util.retry import Retry
 
 from SoftLayer import consts
 from SoftLayer import exceptions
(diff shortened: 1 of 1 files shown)


allmightyspiff:
The retry decorator exists to retry application errors, and the retry adapter exists to retry connection errors. 

Having them retry the same error is certainly not intended for sure, but removing the retry adapter isn't the solution. 


I think removing the `exceptions.TransportError` from the `RETRIABLE ` in https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/decoration.py will fix the 16 retries issue.  If you want to test that out and change your pull request I'll merge it. 

I'm also ok with tuneing down the retries to 2 each, but the retry adapter needs to be kept in.

allmightyspiff:
I've fixed this in #1150 by removing TransportError from one of the retryable errors.

Issue #208, opened 2013-10-09
Ordering a server using a quote id


jravetch:
I do not see a method to order a physical server using a quote id.  Does this exist and if it does how can I use it?

PR #1136, opened 2019-04-18
Ordering doc updates - Quote ordering

allmightyspiff:
Added the following commands:

  *quote              View and Order a quote
  *quote-detail       View a quote
  *quote-list         List all active quotes on an account

Updated some of the ordering related docs as well.


allmightyspiff:
Looks like the verifyOrder problem was because in XML we pass in `''` instead of `None`, so my fix was to just remove any `''` fields from the result we get back from `getRecalculatedOrderContainer`

allmightyspiff:
fixes #208

PR #1143, opened 2019-05-08
Upgrade to prompt_toolkit >= 2

### LINKED ISSUE #1042: prompt_toolkit 2
I'm working on updating the prompt_toolkit package in Fedora to version 2, which is required to update ipython to version 7.  However, some packages, including python2-softlayer and python3-softlayer, still require prompt_toolkit 1.  What is the status of this project migrating to prompt_toolkit 2?

https://python-prompt-toolkit.readthedocs.io/en/stable/pages/upgrading.html

camporter:
Upgrade to prompt_toolkit 2.
* Maintain a session across prompts so that arrow up and down can go through line history.
* Update the documented package in README.
* Return from the completion generator rather than raising a `StopIterator` to stop the generator. This is so that under Python 3.7 with [PEP 479](https://www.python.org/dev/peps/pep-0479/) the iteration ends properly and does not throw ugly errors while typing commands.

Fixes #1042


camporter:
Unfortunately I've had to remove the tests for the shell CLI as I was not able to make them work compatibly across python versions.

Issue #1003, opened 2018-07-16
Networking Feature


allmightyspiff:
- Network topology  view (under summary maybe)
- Network Graphs
  + Bandwidth Histogram
  + save bandwidth image
  + bandwidth summary

allmightyspiff:
docs for 
- slcli summary
- slcli report bandwidth

PR #1150, opened 2019-05-29
Bandwidth usage details for vs and server

allmightyspiff:
#1003 

## New Commands
- vs bandwidth
- hw bandwidth

## Updated Commands
Added bandwidth allotment details
- vs detail 
- hw detail

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,13 @@ matrix:
       env: TOX_ENV=py35
     - python: "3.6"
       env: TOX_ENV=py36
-    - python: "pypy3.7"
+    - python: "3.7"
+      env: TOX_ENV=py37
+    - python: "pypy3.6"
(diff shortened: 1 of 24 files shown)


allmightyspiff:
> Looks good, just there is the littlest difference between VSI or server Bandwidth Detail IN OUT slcli values and Portal Device Details Bandwidth IN OUT values.

Likely due to some rounding on my part, or a slight difference in date windows. Thanks for pointing it out, I'll likely just add a message in the help identifying why they might be different from the portal.

FernandoOjeda:
It look good for me too.

To get the same usage bandwidth result by slcli and control portal the "summary_period" has to be 1800

@click.option('--summary_period', '-p', type=click.INT, default=1800, show_default=True,
              help="300, 600, 1800, 3600, 43200 or 86400 seconds")

PR #1169, opened 2019-09-10
Drop python2.7 support

camporter:
Due to python 2.7 losing official support in January 2020, consider an update that drops support for it.

* Remove six.
* Remove unnecessary `__future__` imports.
* Remove py27 from tox and travisci build.
* Replace pypy with pypy3.
* Add `python_requires` to setup.py and set the minimum to python 3.5.

Other things to consider:
* Fail with a message if library or CLI still used on python 2.

This should probably be considered for a new major version.


allmightyspiff:
Thanks for removing all the old py2 stuff :)

PR #1175, opened 2019-10-03
unit test logs and scale

caberos:
unit test autoscale 
class Logs scale


SoftLayer/fixtures/SoftLayer_Scale_Group.py
```diff
@@ -0,0 +1,54 @@
...
+        "createDate": "2019-10-02T08:34:21-06:00",
+        "id": 3145330,
+        "scaleGroupId": 2245415
+
+    }
+]
+
+scale = []
+
+scaleTo = [{'createDate': '2019-10-02T15:24:56-06:00',
```
allmightyspiff: Can you fix all the space characters between the "-". Will help tests be more accurate for what data actually comes back from the API.

Issue #1095, opened 2019-02-05
Handle missing fixtures better in unit tests.


allmightyspiff:
Given this test

```
    def test_quick(self):
        result = self.run_command(['call-api', 'SoftLayer_Test', 'getTest'])
        self.assert_no_fail(result)
```

Running it will give this result

```
pytest tests/CLI/modules/vs/vs_create_tests.py
================================================================================================= test session starts =================================================================================================
platform darwin -- Python 3.7.0, pytest-3.8.1, py-1.6.0, pluggy-0.7.1
rootdir: /Users/christopher/Code/softlayer-python, inifile: setup.cfg
collected 26 items

tests/CLI/modules/vs/vs_create_tests.py .........................F                                                                                                                                              [100%]

====================================================================================================== FAILURES =======================================================================================================
_____________________________________________________________________________________________ VirtCreateTests.test_quick ______________________________________________________________________________________________
NOTE: Incompatible Exception Representation, displaying natively:

testtools.testresult.real._StringException: Traceback (most recent call last):
  File "/Users/christopher/.virtualenvs/py3/lib/python3.7/site-packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/Users/christopher/.virtualenvs/py3/lib/python3.7/site-packages/urllib3/connectionpool.py", line 384, in _make_request
    six.raise_from(e, None)
  File "<string>", line 2, in raise_from
  File "/Users/christopher/.virtualenvs/py3/lib/python3.7/site-packages/urllib3/connectionpool.py", line 380, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/http/client.py", line 1321, in getresponse
    response.begin()
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/http/client.py", line 296, in begin
    version, status, reason = self._read_status()
  File "/usr/local/Cellar/python/3.7.0/Frameworks/Python.framework/Versions/3.7/lib/python3.7/http/client.py", line 265, in _read_status
    raise RemoteDisconnected("Remote end closed connection without"
http.client.RemoteDisconnected: Remote end closed connection without response

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #1179, opened 2019-10-10
#1095 Handle missing fixtures better in unit tests.

ATGE:


SoftLayer/testing/xmlrpc.py
```diff
@@ -66,6 +66,15 @@ def do_POST(self):
             except UnicodeDecodeError:
                 self.wfile.write(response_body)
 
+        except NotImplementedError as ex:
```
allmightyspiff: Also catch `NameError` here. As that will handle any fixtures that exist, but don't have the method defined in them.

Like this:
```
except (NotImplementedError, NameError) as ex:
```

allmightyspiff:
Fixes #1095

SoftLayer/testing/xmlrpc.py
ATGE: I have added it.

PR #1184, opened 2019-11-14
Fix python misspelling

camporter:
Minor spelling fix and copyright update.


allmightyspiff:
Apparently I need to fix that CDN test. Looks like a slow test run can cause it to fail by the timestamp being 1s off.

Issue #1192, opened 2019-12-01
error in slcli hw detail


scottwn:
**Please triple-check to make sure that you have properly masked out user credentials like usernames, passwords and API keys before submitting your issue**

### Expected Behavior
slcli hw detail should display details of a bare metal when I supply an ID

### Actual Behavior
```
scottwn@MacBook-Pro ~/qiskit-terra> slcli hw detail 1326127
An unexpected error has occured:
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/SoftLayer/CLI/core.py", line 182, in main
    cli.main(**kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/SoftLayer/CLI/hardware/detail.py", line 60, in cli
    bandwidth = hardware.get_bandwidth_allocation(hardware_id)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/SoftLayer/managers/hardware.py", line 701, in get_bandwidth_allocation
[...]

PR #1194, opened 2019-12-05
Fix hardware detail bandwidth allocation.

FernandoOjeda:
Fix hardware detail bandwidth allocation https://github.com/softlayer/softlayer-python/issues/1192.

### DIFF
--- a/SoftLayer/CLI/hardware/detail.py
+++ b/SoftLayer/CLI/hardware/detail.py
@@ -99,12 +99,10 @@ def _bw_table(bw_data):
         allotment = 'N/A'
         if bw_point['type']['alias'] == 'PUBLIC_SERVER_BW':
             bw_type = 'Public'
-            if bw_data.get('allotment') is None:
+            if not bw_data.get('allotment'):
                 allotment = '-'
             else:
                 allotment = utils.lookup(bw_data, 'allotment', 'amount')
-                if allotment is None:
-                    allotment = '-'
 
         table.add_row([bw_type, bw_point['amountIn'], bw_point['amountOut'], allotment])
     return table
--- a/SoftLayer/CLI/hardware/detail.py
+++ b/SoftLayer/CLI/hardware/detail.py
@@ -92,16 +92,19 @@ def cli(env, identifier, passwords, price):
 
 
 def _bw_table(bw_data):
-    """Generates a bandwidth useage table"""
+    """Generates a bandwidth usage table"""
     table = formatting.Table(['Type', 'In GB', 'Out GB', 'Allotment'])
-    for bw_point in bw_data.get('useage'):
+    for bw_point in bw_data.get('usage'):
         bw_type = 'Private'
         allotment = 'N/A'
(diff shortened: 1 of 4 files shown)


SoftLayer/managers/hardware.py
```diff
@@ -697,8 +697,11 @@ def get_bandwidth_allocation(self, instance_id):
         a_mask = "mask[allocation[amount]]"
         allotment = self.client.call('Hardware_Server', 'getBandwidthAllotmentDetail', id=instance_id, mask=a_mask)
         u_mask = "mask[amountIn,amountOut,type]"
-        useage = self.client.call('Hardware_Server', 'getBillingCycleBandwidthUsage', id=instance_id, mask=u_mask)
-        return {'allotment': allotment.get('allocation'), 'useage': useage}
+        usage = self.client.call('Hardware_Server', 'getBillingCycleBandwidthUsage', id=instance_id, mask=u_mask)
+        allotment_bandwidth = None
+        if allotment is not "":
```
allmightyspiff: just do `if allotment` here. The tricky bit is that the XMLRPC endpoint likes to return `None`, and REST will return `''`. Luckily both will return False in python.

SoftLayer/CLI/hardware/detail.py
```diff
@@ -92,16 +92,19 @@ def cli(env, identifier, passwords, price):
...
+    """Generates a bandwidth usage table"""
     table = formatting.Table(['Type', 'In GB', 'Out GB', 'Allotment'])
-    for bw_point in bw_data.get('useage'):
+    for bw_point in bw_data.get('usage'):
         bw_type = 'Private'
         allotment = 'N/A'
         if bw_point['type']['alias'] == 'PUBLIC_SERVER_BW':
             bw_type = 'Public'
-            allotment = utils.lookup(bw_data, 'allotment', 'amount')
-            if allotment is None:
```
allmightyspiff: If a server doesn't have an allotment, what does `utils.lookup(bw_data, 'allotment', 'amount')` return, I'm guessing `''` ? 

If that is the case, you can just change `if allotment is None:` to `if not allotment` to avoid having the nested if statements.

SoftLayer/managers/hardware.py
allmightyspiff: that will also fix the Tox error about comparing to a literal.

caberos: Try using with this  
`if myString in (None, '')`

FernandoOjeda:
Please let me know if the new implementation is ok for you.

SoftLayer/managers/hardware.py
FernandoOjeda: Done.

SoftLayer/CLI/hardware/detail.py
FernandoOjeda: If a server doesn't have an allotment, the  utils.lookup(bw_data, 'allotment', 'amount') returns an error.

PR #1201, opened 2019-12-19
fixing new flask8 errors that show up out of nowhere

allmightyspiff:


allmightyspiff:
fixes some new Flake8 errors that showed up.

Issue #1222, opened 2020-02-06
slcli lb detail <NAME> throws an exception


allmightyspiff:
`SLCLI lb detail <NAME>` should either look up a load balancer by name, or if that isn't feasible, don't print an ugly exception when it couldn't find one.

```
✗ slcli lb detail TEST-307608-dal13.clb.appdomain.cloud
An unexpected error has occured:
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/SoftLayer/CLI/core.py", line 182, in main
    cli.main(**kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/SoftLayer/CLI/loadbal/detail.py", line 16, in cli
    _, lbid = mgr.get_lbaas_uuid_id(identifier)
  File "/usr/local/lib/python2.7/site-packages/SoftLayer/managers/load_balancer.py", line 103, in get_lbaas_uuid_id
    return this_lb['uuid'], this_lb['id']
KeyError: 'uuid'

Feel free to report this error as it is likely a bug:
    https://github.com/softlayer/softlayer-python/issues
The following snippet should be able to reproduce the error
```

PR #1223, opened 2020-02-07
#1222 load balancer by name

ATGE:


SoftLayer/managers/load_balancer.py
```diff
@@ -100,7 +100,20 @@ def get_lbaas_uuid_id(self, identifier):
...
+
+    def get_lbaas_by_address(self, address):
+        """Gets a LBaaS by address.
+
+        :param address: Address of the LBaaS instance
+        """
+        this_lb = {}
+        this_lbs = self.lbaas.getAllObjects()
+        for lbaas in this_lbs:
+            if lbaas.get('address') == address:
```
camporter: Is it possible to lookup the lb by the address using an object filter instead?

SoftLayer/CLI/loadbal/detail.py
```diff
@@ -13,8 +14,15 @@
...
     """Get Load Balancer as a Service details."""
     mgr = SoftLayer.LoadBalancerManager(env.client)
-    _, lbid = mgr.get_lbaas_uuid_id(identifier)
-    this_lb = mgr.get_lb(lbid)
+
+    if utils.valid_domain(identifier):
+        lbaas = mgr.get_lbaas_by_address(identifier)
+        if not lbaas:
+            raise exceptions.CLIAbort("{} address not found".format(identifier))
+        this_lb = mgr.get_lb(lbaas.get('id'))
```
camporter: Do we need to get the lb again after getting it the first time by address?

SoftLayer/managers/load_balancer.py
ATGE: An object filter was the first option that I tried, the object filter works with id, uuid, and name properties, but for some reason when address property is used , it doesn't work.

SoftLayer/CLI/loadbal/detail.py
ATGE: get_lb is adding properties by using a mask, and member health by getLoadBalancerMemberHealth.

allmightyspiff: The reason we have 1 api call to just get the Id/uuid is the LBaaS apis are somewhat inconsistent on when they require an id, and when they require a UUID. So I felt this approach made the lbaas manager simpler and more internally consistent, although I'll likely think of a better solution at some point.

SoftLayer/managers/load_balancer.py
allmightyspiff: apparently `address` isn't a "real" property. I don't want to use `getAllObjects` every time we look up an id.

SoftLayer/managers/load_balancer.py
```diff
@@ -92,15 +92,30 @@ def update_lb_health_monitors(self, uuid, checks):
...
 
-        :param identifier: either the LB Id, or UUID, this function will return both.
+        :param identifier: either the LB Id, UUID or Name, this function will return UUI and LB Id.
         :return (uuid, id):
         """
-        # int objects don't have a len property.
-        if not isinstance(identifier, int) and len(identifier) == 36:
-            this_lb = self.lbaas.getLoadBalancer(identifier, mask="mask[id,uuid]")
+        mask = "mask[id,uuid]"
+        if isinstance(identifier, int):
```
allmightyspiff: Need to also check if the string we got is really just a number. [isdigit](https://docs.python.org/3/library/stdtypes.html#str.isdigit) should be exactly what we need.

`if isinstance(identifier, int) or indentifier.isdigit(): `


```
$ ./slcli -v  lb detail 874328
Calling: SoftLayer_Network_LBaaS_LoadBalancer::getAllObjects(id=None, mask='mask[id,uuid]', filter='{'name': {'operation': '874328'}}', args=(), limit=None, offset=None))
Calling: SoftLayer_Network_LBaaS_LoadBalancer::getObject(id=None, mask='mask[healthMonitors, l7Pools, members, sslCiphers, listeners[defaultPool[healthMonitor,members, sessionAffinity],l7Policies]]', filter='None', args=(), limit=None, offset=None))
SoftLayerAPIError(500): Object does not exist to execute method on. (SoftLayer_Network_LBaaS_LoadBalancer::getObject)

```

SoftLayer/managers/load_balancer.py
```diff
@@ -92,15 +92,30 @@ def update_lb_health_monitors(self, uuid, checks):
...
+        """Gets a LBaaS by name.
+
+        :param name: Name of the LBaaS instance
+        :param mask:
+        :returns: SoftLayer_Network_LBaaS_LoadBalancer or an empty dictionary if the name is not found.
+        """
+        object_filter = {'name': {'operation': name}}
+        this_lbs = self.lbaas.getAllObjects(filter=object_filter, mask=mask)
+
+        return this_lbs[0] if this_lbs else {}
```
allmightyspiff: Lets raise an exception here if we can't find a lbaas instance. This specific service returns a 500 error instead of a 404 which might confuse users.

Something like this I think would be ideal.
```
        if not this_lbs:
            raise exceptions.SoftLayerError("Unable to find LBaaS with name: {}".format(name))
        return this_lbs[0]
```

Issue #1224, opened 2020-02-11
Multiple colon symbols issue


caberos:
Advertency note about using multiple colon symbols in CLI documentation

PR #1225, opened 2020-02-11
add note about using multiple colon symbols

caberos:
https://github.com/softlayer/softlayer-python/issues/1224


docs/cli/hardware.rst
```diff
@@ -43,6 +43,7 @@ Provides some basic functionality to order a server. `slcli order` has a more fu
 
 When setting port speed, use "-1" to indicate best possible configuration. Using 10/100/1000/10000 on a server with a redundant interface may result the interface entering a degraded state. See `setPublicNetworkInterfaceSpeed <http://sldn.softlayer.com/reference/services/SoftLayer_Hardware_Server/setPublicNetworkInterfaceSpeed/>`_ for more information.
 
+**Note :** Using multiple colon symbols can cause an error.
```
allmightyspiff: Can you add the specific error SLCLI returns so its easier to search for. Possibly as a code block example. 

Also, specifically mention that a single `:` in a tag is fine.

Issue #1221, opened 2020-01-30
Version checker


allmightyspiff:
Nice to have feature:

Display current and latest release versions. Ideally only check on certain commands.

`slcli --version`
`slcli -v`
or any command that throws an uncaught exception.

```
slcli --version
Current: slcli (SoftLayer Command-line), version 5.8.0
Latest:   slcli (SoftLayer Command-line), version 5.8.5
```

Something like that.

PR #1226, opened 2020-02-11
#1221 Added version checker

ATGE:


allmightyspiff:
I don't know why that pypy3.5 test is failing, I'll have to figure that out.

SoftLayer/CLI/core.py
```diff
@@ -5,12 +5,14 @@
...
     :license: MIT, see LICENSE for more details.
 """
+import json
 import logging
 import os
 import sys
 import time
 import traceback
 import types
+import urllib3
```
allmightyspiff: use `requests`,  which contains its own json stuff, so you can remove the json import as well.

And add this  constant:
`PROG_NAME = "slcli (SoftLayer Command-line)"`

SoftLayer/CLI/core.py
```diff
@@ -69,6 +71,26 @@ def get_command(self, ctx, name):
...
+    try:
+        str_result = http.request('GET', 'https://pypi.python.org/pypi/softlayer/json')
+        dic_result = json.loads(str_result.data.decode('utf-8'))
+        latest = dic_result['info']['version']
+    except Exception:
+        latest = '%(version)s'
+    return latest
+
+
+def get_version_message():
```
allmightyspiff: This will actually call `get_latest_version()` for EVERY command run. So lets do the following:

```python
def get_version_message(ctx, param, value):
    if not value or ctx.resilient_parsing:
        return

    current = SoftLayer.consts.VERSION
    latest = get_latest_version()
    click.secho("Current: {prog} {current}\nLatest:  {prog} v{latest}".format(
        prog=PROG_NAME, current=current, latest=latest))
    ctx.exit()
```

SoftLayer/CLI/core.py
```diff
@@ -103,7 +125,8 @@ def get_command(self, ctx, name):
               is_flag=True,
               required=False,
               help="Use demo data instead of actually making API calls")
-@click.version_option(prog_name="slcli (SoftLayer Command-line)")
+@click.version_option(prog_name="slcli (SoftLayer Command-line)",
```
allmightyspiff: Doing this will allow us to only query for the latest version only when someone actually runs `--version`. 
```
@click.option('--version', is_flag=True, expose_value=False, is_eager=True, callback=get_version,
              help="Show version information.")
```

PR #1228, opened 2020-02-17
Support Ordering Dependent Duplicate storage volumes

kremlin-:
supports order dep. dupe. volumes for file/block

### DIFF
--- a/SoftLayer/CLI/block/duplicate.py
+++ b/SoftLayer/CLI/block/duplicate.py
@@ -54,9 +54,15 @@ CONTEXT_SETTINGS = {'token_normalize_func': lambda x: x.upper()}
               type=click.Choice(['hourly', 'monthly']),
               default='monthly',
               help="Optional parameter for Billing rate (default to monthly)")
+@click.option('--dependent-duplicate',
+              type=click.BOOL,
+              default=False,
+              help='Whether or not this duplicate will be a dependent duplicate '
+                    'of the origin volume (default to false)')
 @environment.pass_env
 def cli(env, origin_volume_id, origin_snapshot_id, duplicate_size,
-        duplicate_iops, duplicate_tier, duplicate_snapshot_size, billing):
+        duplicate_iops, duplicate_tier, duplicate_snapshot_size, billing,
+        dependent_duplicate):
     """Order a duplicate block storage volume."""
     block_manager = SoftLayer.BlockStorageManager(env.client)
 
@@ -75,7 +81,8 @@ def cli(env, origin_volume_id, origin_snapshot_id, duplicate_size,
             duplicate_iops=duplicate_iops,
             duplicate_tier_level=duplicate_tier,
             duplicate_snapshot_size=duplicate_snapshot_size,
-            hourly_billing_flag=hourly_billing_flag
+            hourly_billing_flag=hourly_billing_flag,
+            dependent_duplicate=dependent_duplicate
         )
     except ValueError as ex:
         raise exceptions.ArgumentError(str(ex))
--- a/SoftLayer/CLI/block/duplicate.py
+++ b/SoftLayer/CLI/block/duplicate.py
@@ -58,7 +58,7 @@ CONTEXT_SETTINGS = {'token_normalize_func': lambda x: x.upper()}
               type=click.BOOL,
               default=False,
               help='Whether or not this duplicate will be a dependent duplicate '
-                    'of the origin volume (default to false)')
+                   'of the origin volume (default to false)')
 @environment.pass_env
 def cli(env, origin_volume_id, origin_snapshot_id, duplicate_size,
         duplicate_iops, duplicate_tier, duplicate_snapshot_size, billing,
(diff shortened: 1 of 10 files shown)


SoftLayer/CLI/block/duplicate.py
```diff
@@ -54,9 +54,15 @@
               type=click.Choice(['hourly', 'monthly']),
               default='monthly',
               help="Optional parameter for Billing rate (default to monthly)")
+@click.option('--dependent-duplicate',
+              type=click.BOOL,
+              default=False,
+              help='Whether or not this duplicate will be a dependent duplicate '
+                    'of the origin volume (default to false)')
```
allmightyspiff: `show_default=True` option will print out the default in the help message, so you don't have to specify here.

https://click.palletsprojects.com/en/7.x/options/

SoftLayer/CLI/file/duplicate.py
```diff
@@ -52,9 +52,16 @@
               type=click.Choice(['hourly', 'monthly']),
               default='monthly',
               help="Optional parameter for Billing rate (default to monthly)")
+@click.option('--dependent-duplicate',
+              type=click.BOOL,
+              default=False,
+              help='Whether or not this duplicate will be a dependent duplicate'
+                    'of the origin volume (default to false)')
```
allmightyspiff: Use `show_default=True`

SoftLayer/managers/block.py
```diff
@@ -348,6 +351,9 @@ def order_duplicate_volume(self, origin_volume_id, origin_snapshot_id=None,
         if origin_snapshot_id is not None:
             order['duplicateOriginSnapshotId'] = origin_snapshot_id
 
+        if dependent_duplicate:
+            order['isDependentDuplicateFlag'] = 1
```
allmightyspiff: Would be simpler to just have `order['isDependentDuplicateFlag'] = dependent_duplicate` and remove the if statement

SoftLayer/managers/file.py
```diff
@@ -289,6 +292,9 @@ def order_duplicate_volume(self, origin_volume_id, origin_snapshot_id=None,
         if origin_snapshot_id is not None:
             order['duplicateOriginSnapshotId'] = origin_snapshot_id
 
+        if dependent_duplicate:
+            order['isDependentDuplicateFlag'] = 1
```
allmightyspiff: Would be simpler to just have order['isDependentDuplicateFlag'] = dependent_duplicate and remove the if statement

SoftLayer/managers/storage_utils.py
```diff
@@ -903,6 +906,9 @@ def prepare_duplicate_order_object(manager, origin_volume, iops, tier,
     if volume_is_performance:
         duplicate_order['iops'] = iops
 
+    if dependent_duplicate:
+        duplicate_order['isDependentDuplicateFlag'] = 1
```
allmightyspiff: Would be simpler to just have order['isDependentDuplicateFlag'] = dependent_duplicate and remove the if statement

SoftLayer/managers/block.py
```diff
@@ -321,6 +322,8 @@ def order_duplicate_volume(self, origin_volume_id, origin_snapshot_id=None,
         :param duplicate_snapshot_size: Snapshot space size for the duplicate
         :param hourly_billing_flag: Billing type, monthly (False)
             or hourly (True), default to monthly.
+        :param dependent_duplicate: Duplicate type, normal (False) or dependent
```
allmightyspiff: you can use 120 character lines.

SoftLayer/CLI/block/duplicate.py
kremlin-: as we discussed, this is not possible

SoftLayer/managers/block.py
kremlin-: fixed in 7a651b0

SoftLayer/managers/storage_utils.py
kremlin-: as we discussed, this isn't possible

SoftLayer/managers/file.py
kremlin-: as we discussed, this isn't possible

SoftLayer/managers/block.py
kremlin-: as we discussed, this isn't possible

SoftLayer/CLI/block/duplicate.py
kremlin-: fixed in e424ce1

SoftLayer/CLI/file/duplicate.py
kremlin-: fixed in e424ce1

kremlin-:
> Use the `show_default=True` click CLI option instead of having the default be in the help message.
> 
> Instead of using the `if dependent_duplicate` check, just set the isDependentDuplicateFlag to whatever depend_duplicate is. Since it has a default value of 0/False, that should be fine.

this should all be fixed now

PR #1231, opened 2020-02-20
bring in convert/refresh funcs

kremlin-:


SoftLayer/managers/block.py
```diff
@@ -616,3 +616,20 @@ def create_or_update_lun_id(self, volume_id, lun_id):
         """
         return self.client.call('Network_Storage', 'createOrUpdateLunId',
                                 lun_id, id=volume_id)
+
+    def refresh_dep_dupe(self, volume_id, snapshot_id):
```
allmightyspiff: Since these 2 commands are literally the same for file/block, just put them in the storage_utils.py file so we don't have as much duplicated code in these 2 managers than we already do.

SoftLayer/managers/block.py
```diff
@@ -616,3 +616,20 @@ def create_or_update_lun_id(self, volume_id, lun_id):
...
+        """"Refresh a dependent duplicate volume with a snapshot from its parent.
+
+        :param integer volume_id: The id of the volume
+        :param integer snapshot_id: The id of the snapshot
+        """
+        return self.client.call('Network_Storage', 'refreshDependentDuplicate',
```
allmightyspiff: make this 1 line.

SoftLayer/managers/block.py
```diff
@@ -616,3 +616,20 @@ def create_or_update_lun_id(self, volume_id, lun_id):
...
+                                snapshot_id, id=volume_id)
+
+    def convert_dep_dupe(self, volume_id):
+        """Convert a dependent duplicate volume to an indepdent volume.
+
+        :param integer volume_id: The id of the volume.
+        """
+        return self.client.call('Network_Storage', 'convertCloneDependentToIndependent',
```
allmightyspiff: make this 1 line

allmightyspiff:
I've refactored the file/block managers, which will likely conflict with your changes. 

When you get a chance, you might want to pull in the changes from https://github.com/softlayer/softlayer-python/pull/1235 (or just wait until its merged to master). 

The practical effect is that you'll just put your file/block manager new functions into the new `storage` manager since they are the exact same between file and block.

SoftLayer/managers/block.py
kremlin-: fixed, storage.py manager though, not utils

SoftLayer/managers/block.py
kremlin-: 👍

SoftLayer/managers/block.py
kremlin-: 👍

Issue #1229, opened 2020-02-17
slcli [block|file] volume-detail <bad name> throws exception


allmightyspiff:
```
slcli -v  file volume-detail SL01SEL307608-11
Calling: SoftLayer_Network_Storage::getObject(id=SL01SEL307608-11, mask='id,username,password,capacityGb,bytesUsed,snapshotCapacityGb,parentVolume.snapshotSizeBytes,storageType.keyName,serviceResource.datacenter[name],serviceResourceBackendIpAddress,fileNetworkMountAddress,storageTierLevel,provisionedIops,lunId,originalVolumeName,originalSnapshotName,originalVolumeSize,activeTransactionCount,activeTransactions.transactionStatus[friendlyName],replicationPartnerCount,replicationStatus,replicationPartners[id,username,serviceResourceBackendIpAddress,serviceResource[datacenter[name]],replicationSchedule[type[keyname]]]', filter='None', args=(), limit=None, offset=None))
An unexpected error has occured:
Traceback (most recent call last):
  File "c:\users\allmi\source\py37\lib\site-packages\SoftLayer\CLI\core.py", line 182, in main
    cli.main(**kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\click\decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\SoftLayer\CLI\file\detail.py", line 24, in cli
    storage_type = file_volume['storageType']['keyName'].split('_').pop(0)
[...]

PR #1232, opened 2020-02-20
Fix the block and file storage detail.

FernandoOjeda:
Fix block and file storage detail https://github.com/softlayer/softlayer-python/issues/1229.
This works for block and file storage with id and name identifier.

**Valid name:**
slcli -v file volume-detail SL01SEV30_1

**Valid id**
slcli -v file volume-detail 11111

**Bad name**
slcli -v file volume-detail test
Error: Unable to find File Storage 'test'

### DIFF
--- a/SoftLayer/CLI/block/detail.py
+++ b/SoftLayer/CLI/block/detail.py
@@ -5,7 +5,18 @@ import click
 import SoftLayer
 from SoftLayer.CLI import environment
 from SoftLayer.CLI import formatting
+from SoftLayer.CLI import helpers
 from SoftLayer import utils
+
+
+def get_block_volume_id(volume_id, block_manager):
+    storage_list = block_manager.list_block_volumes()
+    for storage in storage_list:
+        if volume_id == storage['username']:
+            volume_id = storage['id']
+            break
+
+    return volume_id
 
 
 @click.command()
@@ -14,7 +25,9 @@ from SoftLayer import utils
 def cli(env, volume_id):
     """Display details for a specified volume."""
     block_manager = SoftLayer.BlockStorageManager(env.client)
-    block_volume = block_manager.get_block_volume_details(volume_id)
+    volume_id = get_block_volume_id(volume_id, block_manager)
+    block_volume_id = helpers.resolve_id(block_manager.resolve_ids, volume_id, 'Block Volume')
+    block_volume = block_manager.get_block_volume_details(block_volume_id)
     block_volume = utils.NestedDict(block_volume)
 
     table = formatting.KeyValueTable(['Name', 'Value'])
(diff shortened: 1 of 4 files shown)


SoftLayer/CLI/block/detail.py
```diff
@@ -5,16 +5,35 @@
 import SoftLayer
 from SoftLayer.CLI import environment
 from SoftLayer.CLI import formatting
+from SoftLayer.CLI import helpers
 from SoftLayer import utils
 
 
+def get_block_volume_id(volume_id, block_manager):
```
allmightyspiff: lets move this to the block and file managers as `_get_ids_from_username(username)`

Then add it as a resolver to the manager itself.

```python
class BlockStorageManager(utils.IdentifierMixin, object):
    def __init__(self, client):
        self.configuration = {}
        self.client = client
        self.resolvers = [self._get_ids_from_username]
```

SoftLayer/CLI/block/detail.py
```diff
@@ -5,16 +5,35 @@
...
+    """Returns the volume id.
+
+    :param volume_id: ID of volume.
+    :param block_manager: Block Storage Manager.
+    :return: Returns the volume id.
+    """
+    storage_list = block_manager.list_block_volumes()
```
allmightyspiff: `list_block_volumes()` takes in a username parameter to filter by that, which will let you only get the volume back that matches the username. Also pass in a mask of just `id` help reduce API call returned data.

SoftLayer/CLI/block/detail.py
```diff
@@ -5,16 +5,35 @@
...
+
+
 @click.command()
 @click.argument('volume_id')
 @environment.pass_env
 def cli(env, volume_id):
     """Display details for a specified volume."""
     block_manager = SoftLayer.BlockStorageManager(env.client)
-    block_volume = block_manager.get_block_volume_details(volume_id)
+    volume_id = get_block_volume_id(volume_id, block_manager)
```
allmightyspiff: since were moving get_block_volume_id to the block manager's resolvers feature, we can just remove this line and the rest should work.

tests/CLI/modules/file_tests.py
```diff
@@ -163,6 +163,49 @@ def test_volume_detail(self):
             ]
         }, json.loads(result.output))
 
+    def test_volume_detail_name_identifier(self):
+        result = self.run_command(['file', 'volume-detail', 'user'])
+
+        self.assert_no_fail(result)
+        self.assertEqual({
```
allmightyspiff: We don't really need to test the WHOLE output is equal, that will make this test break anytime we change the fixture.

Instead you can do something like this to test a few fields of the output

```python
    self.assert_called_with('SoftLayer_Network_Storage', 'getObject')
    self.assertIn('username', result.output)
    self.assertIn('READHEAVY_TIER', result.output)
```
or whatever other strings you want to look for to make sure it called the correct data.

allmightyspiff:
I've added your changes to #1235

Issue #887, opened 2017-10-19
SLCLI - Response shows additional new lines (\n) in ticket details 


acamacho82:
A softlayer customer reported the following:

> "May I kindly ask you to correct an issue with slcli version 5.2.14 which displaying hostnames in "slcli ticket detail  [id]" with unexpected line breaks in comparison to the web portal version.
> It is very hard to automate and parse those content with this annoying displayed output."

I reproduced the issue, see attached image.

I executed **"slcli --format=json ticket detail [id]"** and comparing with REST, the SLCLI returns the description with additional "new lines" ( \n ).
 
### Expected Behavior
Ticket details should be displayed like in portal page, without extra new lines \n

### Actual Behavior
Ticket details isn't displayed like in portal page because it has extra new lines.
 
### Environment Information
Operating System: Windows
slcli version: 5.2.12  and 5.2.14
![slcliissue](https://user-images.githubusercontent.com/6937170/31793541-837fe95a-b4ed-11e7-88c6-4012ae77aaf8.png)

PR #1246, opened 2020-03-24
fix the issue 887

caberos:
https://github.com/softlayer/softlayer-python/issues/887
- fixed too many newlins in tickets

### DIFF
--- a/SoftLayer/CLI/ticket/__init__.py
+++ b/SoftLayer/CLI/ticket/__init__.py
@@ -17,7 +17,7 @@ PRIORITY_MAP = [
 ]
 
 
-def get_ticket_results(mgr, ticket_id, is_json=False, update_count=1):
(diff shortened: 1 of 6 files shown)


SoftLayer/CLI/ticket/__init__.py
```diff
@@ -64,6 +64,8 @@ def get_ticket_results(mgr, ticket_id, update_count=1):
 
         # NOTE(kmcdonald): Windows new-line characters need to be stripped out
         wrapped_entry += click.wrap_text(update['entry'].replace('\r', ''))
+        if is_json:
```
allmightyspiff, line 67: so `click.wrapp_text` is where the extra new lines are coming from. 

So lets do this instead of using regex:

```python
if is_json:
    wrapped_entry += updated['entry']
else:
    wrapped_entry += click.wrap_text(update['entry'].replace('\r', ''), width=120)
```

PR #1254, opened 2020-04-06
added account billing-items/item-details/cancel-item commands

allmightyspiff:
- slcli account billing-items
- slcli account item-detail
- slcli account cancel-item

Instead of doing a report, just basically reproduced https://cloud.ibm.com/billing/billing-items
with item-detail and cancel-item commands as needed.


allmightyspiff:
TODO: Add documentation.

Issue #1266, opened 2020-04-29
Ticket upload doesn't work in REST


allmightyspiff:
```
$ slcli ticket upload --path ./out.txt --name out.txt  108158052
An unexpected error has occured:
Traceback (most recent call last):
  File "c:\users\allmi\source\py37\lib\site-packages\SoftLayer\CLI\core.py", line 207, in main
    cli.main(**kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\click\decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\SoftLayer\CLI\ticket\upload.py", line 36, in cli
    file_name=name)
  File "c:\users\allmi\source\py37\lib\site-packages\SoftLayer\managers\ticket.py", line 104, in upload_attachment
    return self.ticket.addAttachedFile(file_object, id=ticket_id)
  File "c:\users\allmi\source\py37\lib\site-packages\SoftLayer\API.py", line 388, in call_handler
    return self(name, *args, **kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\SoftLayer\API.py", line 356, in call
    return self.client.call(self.name, name, *args, **kwargs)
  File "c:\users\allmi\source\py37\lib\site-packages\SoftLayer\API.py", line 264, in call
    return self.transport(request)
  File "c:\users\allmi\source\py37\lib\site-packages\SoftLayer\transports.py", line 461, in __call__
    call.result = self.transport(call)
[...]

PR #1267, opened 2020-04-29
Fix ticket upload to be able to run with rest and xmlrpc endpoint.

FernandoOjeda:
Fix ticket upload to be able to run with rest and xmlrpc endpoint https://github.com/softlayer/softlayer-python/issues/1266.

### DIFF
--- a/SoftLayer/managers/ticket.py
+++ b/SoftLayer/managers/ticket.py
@@ -5,6 +5,9 @@
 
     :license: MIT, see LICENSE for more details.
 """
+import base64
+from json import dumps
+
 from SoftLayer import utils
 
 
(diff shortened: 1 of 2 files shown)


allmightyspiff:
After testing this a bit, looks like there were a few problems with this solution.

1. It breaks the XMLRPC call because the XMLRPC type will be <string> instead of <base64>, and I guess the API doesn't deal with that. The files I uploaded with XML were all broken.
2. Doesn't really fix any other instances that might pop up of json not being able to encode byte strings.

I added a custom JSON encoder to the transports in #1269 that should fix this. 

Thanks for taking a look at this so quickly though.

Issue #1283, opened 2020-06-16
Subnet Editing (tagging)


allmightyspiff:
Add new command `slcli subnet edit`

Should allow users to tag subnets ( with the newly added https://sldn.softlayer.com/reference/services/SoftLayer_Network_Subnet/setTags/ )
Should allow users to set subnet notes

allmightyspiff:
`slcli subnet detail` should print out any tags a subnet has, along with its note.

PR #1286, opened 2020-06-19
Subnet Editing (tagging) #1283

ATGE:


SoftLayer/managers/network.py
```diff
@@ -226,6 +237,20 @@ def cancel_subnet(self, subnet_id):
         billing_id = subnet['billingItem']['id']
         return self.client['Billing_Item'].cancelService(id=billing_id)
 
+    def set_tags_subnet(self, subnet_id, tags):
+        """Tag a subnet by passing in one or more tags separated by a comma.
+
+        :param int subnet_id: The ID of the subnet.
```
allmightyspiff, line 243: Add the `tags` param documentation

SoftLayer/managers/network.py
```diff
@@ -226,6 +237,20 @@ def cancel_subnet(self, subnet_id):
...
+        """
+        return self.subnet.setTags(tags, id=subnet_id)
+
+    def edit_note_subnet(self, subnet_id, note):
+        """Edit the note for this subnet.
+
+        :param int subnet_id: The ID of the subnet.
```
allmightyspiff, line 250: Add the `tags` param documentation

SoftLayer/managers/network.py
ATGE: done

SoftLayer/managers/network.py
ATGE: done

Issue #1289, opened 2020-06-29
ordering not able to lookup RHEL prices properly


allmightyspiff:
```
 slcli order place SINGLE_E31270_V3_4_DRIVES DALLAS13 INTEL_SINGLE_XEON_1270_3_50 RAM_32_GB_DDR3_1333_REG_2 OS_RHEL_7_1_64_BIT DISK_CONTROLLER_NONRAID HARD_DRIVE_1_00_TB_SATA_2 BANDWIDTH_500_GB 1_GBPS_PUBLIC_PRIVATE_NETWORK_UPLINKS 1_IP_ADDRESS REBOOT_KVM_OVER_IP NESSUS_VULNERABILITY_ASSESSMENT_REPORTING NOTIFICATION_EMAIL_AND_TICKET MONITORING_HOST_PING_AND_TCP_SERVICE UNLIMITED_SSL_VPN_USERS_1_PPTP_VPN_USER_PER_ACCOUNT AUTOMATED_NOTIFICATION --complex-type SoftLayer_Container_Product_Order_Hardware_Server --extras '{"hardware":[{"domain":"test-cgallo.com","hostname":"testOrderCANCEL"}]}'  --billing monthly --verify

Calling: SoftLayer_Product_Order::verifyOrder(id=None, mask='', filter='None', args=({'orderContainers': [{'hardware': [{'domain': 'test-cgallo.com', 'hostname': 'testOrderCANCEL'}], 'packageId': 257, 'quantity': 1, 'location': 1854895, 'useHourlyPricing': False, 'complexType': 'SoftLayer_Container_Product_Order_Hardware_Server', 'prices': [{'id': 49515}, {'id': 49415}, {'id': None}, {'id': 876}, {'id': 49759}, {'id': 50357}, {'id': 274}, {'id': 21}, {'id': 906}, {'id': 418}, {'id': 57}, {'id': 56}, {'id': 420}, {'id': 58}]}]},), limit=None, offset=None))
SoftLayerAPIError(500): Price does not have an id.

```

Using a different OS seems to work.

```
[...]

PR #1290, opened 2020-06-30
Add functionality to order a hw for capacityRestrictionType PROCCESOR.

FernandoOjeda:
Add functionality to order a hw for capacityRestrictionType PROCCESOR https://github.com/softlayer/softlayer-python/issues/1289.

### DIFF
--- a/SoftLayer/managers/ordering.py
+++ b/SoftLayer/managers/ordering.py
@@ -396,7 +396,7 @@ class OrderingManager(object):
                 capacity_min = int(price.get('capacityRestrictionMinimum', -1))
                 capacity_max = int(price.get('capacityRestrictionMaximum', -1))
                 # return first match if no restirction, or no core to check
-                if capacity_min == -1 or core is None:
+                if capacity_min == -1 or core is None or "PROCESSOR" in price.get("capacityRestrictionType"):
                     price_id = price['id']
                 # this check is mostly to work nicely with preset configs
                 elif capacity_min <= int(core) <= capacity_max:
--- a/SoftLayer/managers/ordering.py
+++ b/SoftLayer/managers/ordering.py
@@ -393,16 +393,23 @@ class OrderingManager(object):
         price_id = None
         for price in prices:
             if not price['locationGroupId']:
-                capacity_min = int(price.get('capacityRestrictionMinimum', -1))
-                capacity_max = int(price.get('capacityRestrictionMaximum', -1))
-                # return first match if no restirction, or no core to check
-                if capacity_min == -1 or core is None or "PROCESSOR" in price.get("capacityRestrictionType"):
-                    price_id = price['id']
-                # this check is mostly to work nicely with preset configs
-                elif capacity_min <= int(core) <= capacity_max:
-                    if "STORAGE" in price.get("capacityRestrictionType") or "CORE" in price.get(
-                            "capacityRestrictionType"):
+                restriction = price.get('capacityRestrictionType', False)
+                # There is a price restriction. Make sure the price is within the restriction
+                if restriction and core is not None:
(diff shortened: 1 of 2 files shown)


FernandoOjeda:
I did not find any relationship between the os and the other items such as cpu.

e.g.
https://sldn.softlayer.com/reference/services/SoftLayer_Product_Package/getItemPrices/

`[
{
        "currentPriceFlag": null,
        "id": 49515,
        "itemId": 4578,
        "laborFee": "0",
        "locationGroupId": null,
        "recurringFee": "75",
        "tierMinimumThreshold": null,
        "item": {
            "capacity": "3.5",
            "description": "Single Intel Xeon E3-1270 v3 (4 Cores, 3.50 GHz)",
            "id": 4578,
            "keyName": "INTEL_SINGLE_XEON_1270_3_50",
        }
    },
{
        "currentPriceFlag": null,
        "id": 48999,
        "itemId": 6125,
        "laborFee": "0",
        "locationGroupId": null,
        "recurringFee": "45",
        "setupFee": "0",
        "sort": 1,
        "capacityRestrictionMaximum": "1",
        "capacityRestrictionMinimum": "1",
        "capacityRestrictionType": "PROCESSOR",
        "item": {
            "description": "Red Hat Enterprise Linux 7.x (64 bit)  (per-processor licensing)",
            "id": 6125,
            "keyName": "OS_RHEL_7_1_64_BIT",
        }
    },
]`

allmightyspiff:
So, I don't think users have the option to change the number of Processors in a package, so just skipping `PROCESSOR` restricted items should be technically fine, I think refactoring this method a bit would help make it more readable.


Here is what I'm thinking.


```python
    @staticmethod
    def get_item_price_id(core, prices):
        """get item price id"""
        price_id = None
        for price in prices:
            # This is the default location group then
            if not price['locationGroupId']:
                restriction = price.get('capacityRestrictionType', False)
                # There is a price restriction. Make sure the price is within the restriction
                if restriction:
                    capacity_min = int(price.get('capacityRestrictionMinimum', -1))
                    capacity_max = int(price.get('capacityRestrictionMaximum', -1))
                    if restriction == "STORAGE":
                        if capacity_min <= int(core) <= capacity_max:
                            price_id = price['id'] 
                    if restriction == "CORE":
                        if capacity_min <= int(core) <= capacity_max:
                            price_id = price['id'] 
                    if restriction == "PROCESSOR":
                        price_id = price['id']
                # No price restrictions
                else:
                    price_id = price['id']

        return price_id
```
This way we can handle each restriction a little different if needed. Still not great looking, but what can you do.


I thought about just looking up the [processor capacity](https://sldn.softlayer.com/reference/datatypes/SoftLayer_Product_Item/#totalprocessorcapacity) but that seems like a bit more work than just assuming the package will only have 1 type of processor, which so far seems to be the case.

FernandoOjeda:
yep, I am agree refactoring this method using different ` if` for each restriction. It looks good.

I changing the `restriction == "STORAGE"` to `"STORAGE" in restriction` because there are other "capacityRestrictionType" for storage such as STORAGE_SPACE, STORAGE_TIER_LEVEL.

It is ok the implementation below or I change the rest of `if` statements the same that "STORAGE" in restriction.

`                    if "STORAGE" in restriction:
                        if capacity_min <= int(core) <= capacity_max:
                            price_id = price['id']
                    if restriction == "CORE":
                        if capacity_min <= int(core) <= capacity_max:
                            price_id = price['id']
                    if restriction == "PROCESSOR":
                        price_id = price['id']`

allmightyspiff:
> It is ok the implementation below or I change the rest of if statements the same that "STORAGE" in restriction.

Yeah that should be fine.

FernandoOjeda:
Done.

PR #1291, opened 2020-07-01
Removed overly specific reference to Block storage

JonWoodlief:
I found a specific reference to block storage in code shared between file and block. reworded it to refer to either block or storage volumes

### DIFF
--- a/SoftLayer/managers/storage.py
+++ b/SoftLayer/managers/storage.py
@@ -396,7 +396,7 @@ class StorageManager(utils.IdentifierMixin, object):
         return self.client.call('Network_Storage', 'failbackFromReplicant', id=volume_id)
 
     def cancel_volume(self, volume_id, reason='No longer needed', immediate=False):
-        """Cancels the given block storage volume.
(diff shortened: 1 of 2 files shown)


SoftLayer/managers/storage.py
```diff
@@ -416,7 +416,7 @@ def cancel_volume(self, volume_id, reason='No longer needed', immediate=False):
         return self.client.call('SoftLayer_Billing_Item', 'cancelItem', immediate, True, reason, id=billing_item_id)
 
     def refresh_dep_dupe(self, volume_id, snapshot_id):
-        """"Refresh a dependent duplicate volume with a snapshot from its parent.
+        """"Refresh a dependent duplicate volume with a snapshot from its parent.Block Storage was already cancelled
```
allmightyspiff, line 419: Add a space after the `.`

JonWoodlief:
Good catch on the dep_dupe error- didn't mean to edit that actually, must have been an errant keystroke.


I edited the test case, looks like everything is passing now

Issue #1292, opened 2020-07-01
Update  slcli -v hw create-options to use operatingSystemReferenceCode


allmightyspiff:
```
 slcli -v hw create-options

CentOS 6.5-32                                                              OS_CENTOS_6_X_32_BIT
CentOS 6.9-64                                                              OS_CENTOS_6_X_64_BIT
CentOS 6.5-64 Minimal for Bare Metal                                       OS_CENTOS_6_X_MINIMAL_64_BIT
CentOS 7.6-64                                                              OS_CENTOS_7_X_64_BIT
CentOS 7.0-64 Minimal for Bare Metal                                       OS_CENTOS_7_X_MINIMAL_64_BIT_2

```

those options won't work for ordering, so we need to output the OS reference code ("referenceCode": "CENTOS_7_64" ) for example

PR #1293, opened 2020-07-03
add system operation referenceCode in create-option

caberos:
https://github.com/softlayer/softlayer-python/issues/1292


allmightyspiff:
LGTM

PR #1296, opened 2020-07-10
`slcli hw create` updates

allmightyspiff:
Added networking options to `slcli hw create-options`
Refactored `slcli hw create` to use the ordering manager
Added `--network` option to `slcli hw create` for more granular network choices.
deprecated `--port-speed` and `--no-public` . They still work for now, but will be removed in a future release.


allmightyspiff:
@atge I agree with your comment, I created https://github.com/softlayer/softlayer-python/issues/1299 to add something like that feature in the future. I tried figuring out a better way of showing prices/availability/restrictions but it ended up being too complicated.

Issue #1295, opened 2020-07-10
VS Updates


allmightyspiff:
`slcli vs upgrade` should allow for adding additional disks to the guest

~`slcli vs upgrade` should also deprecated CPU/Memory options in favor of changing the guests flavor.~

Essentially, replicate the resize page

![image](https://user-images.githubusercontent.com/7408017/87115244-50eeb600-c238-11ea-89e7-e6f8f402bda2.png)

PR #1308, opened 2020-07-24
vs upgrade disk and add new disk

caberos:
https://github.com/softlayer/softlayer-python/issues/1295

### DIFF
--- a/SoftLayer/CLI/virt/upgrade.py
+++ b/SoftLayer/CLI/virt/upgrade.py
@@ -1,8 +1,6 @@
 """Upgrade a virtual server."""
 # :license: MIT, see LICENSE for more details.
 
-import json
-
 import click
 
 import SoftLayer
@@ -22,18 +20,20 @@ completed. However for Network, no reboot is required.""")
               help="CPU core will be on a dedicated host server.")
 @click.option('--memory', type=virt.MEM_TYPE, help="Memory in megabytes")
 @click.option('--network', type=click.INT, help="Network port speed in Mbps")
-@click.option('--add', type=click.INT, required=False, help="add Hard disk in GB")
-@click.option('--disk', nargs=1, help="update the number and capacity in GB Hard disk, E.G {'number':2,'capacity':100}")
+@click.option('--add-disk', type=click.INT, multiple=True, required=False, help="add Hard disk in GB")
+@click.option('--resize-disk', nargs=2, multiple=True, type=(int, int),
+              help="Update disk number to size in GB. --resize-disk 250 2 ")
 @click.option('--flavor', type=click.STRING,
               help="Flavor keyName\nDo not use --memory, --cpu or --private, if you are using flavors")
 @environment.pass_env
-def cli(env, identifier, cpu, private, memory, network, flavor, disk, add):
+def cli(env, identifier, cpu, private, memory, network, flavor, add_disk, resize_disk):
     """Upgrade a virtual server."""
 
     vsi = SoftLayer.VSManager(env.client)
 
-    if not any([cpu, memory, network, flavor, disk, add]):
-        raise exceptions.ArgumentError("Must provide [--cpu], [--memory], [--network], or [--flavor] to upgrade")
+    if not any([cpu, memory, network, flavor, resize_disk, add_disk]):
+        raise exceptions.ArgumentError("Must provide [--cpu],"
+                                       " [--memory], [--network], [--flavor], [--resize-disk], or [--add] to upgrade")
 
     if private and not cpu:
         raise exceptions.ArgumentError("Must specify [--cpu] when using [--private]")
@@ -44,9 +44,18 @@ def cli(env, identifier, cpu, private, memory, network, flavor, disk, add):
 
     if memory:
         memory = int(memory / 1024)
-    if disk is not None:
-        disk = json.loads(disk)
+    if resize_disk:
+        disk_json = list()
+        for guest_disk in resize_disk:
+            disks = {'capacity': guest_disk[0], 'number': guest_disk[1]}
+            disk_json.append(disks)
+
+    elif add_disk:
+        disk_json = list()
+        for guest_disk in add_disk:
+            disks = {'capacity': guest_disk, 'number': -1}
+            disk_json.append(disks)
 
     if not vsi.upgrade(vs_id, cpus=cpu, memory=memory, nic_speed=network, public=not private, preset=flavor,
-                       disk=disk, add=add):
+                       disk=disk_json):
         raise exceptions.CLIAbort('VS Upgrade Failed')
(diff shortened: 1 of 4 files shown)


SoftLayer/CLI/virt/upgrade.py
```diff
@@ -20,15 +22,17 @@
...
 @environment.pass_env
-def cli(env, identifier, cpu, private, memory, network, flavor):
+def cli(env, identifier, cpu, private, memory, network, flavor, disk, add):
     """Upgrade a virtual server."""
 
     vsi = SoftLayer.VSManager(env.client)
 
-    if not any([cpu, memory, network, flavor]):
+    if not any([cpu, memory, network, flavor, disk, add]):
         raise exceptions.ArgumentError("Must provide [--cpu], [--memory], [--network], or [--flavor] to upgrade")
```
allmightyspiff, line 36: Update this exception to be "No upgrade options provided"

SoftLayer/managers/vs.py
```diff
@@ -874,9 +879,30 @@ def upgrade(self, instance_id, cpus=None, memory=None, nic_speed=None, public=Tr
...
-            prices.append({'id': price_id})
-        order['prices'] = prices
-
+            if disk is not None:
+                category = {'categories': [{
+                    'categoryCode': 'guest_disk' + str(disk.get('number')),
+                    'complexType': "SoftLayer_Product_Item_Category"
+                }], 'complexType': 'SoftLayer_Product_Item_Price'}
+                prices.append(category)
+                prices[0]['id'] = price_id
```
allmightyspiff, line 888: You should set the id in the `category` object. Since its not certain that prices[0] will be the disk price.

```python
                category = {'categories': [{
                    'categoryCode': 'guest_disk' + str(disk.get('number')),
                    'complexType': "SoftLayer_Product_Item_Category"
                }], 'complexType': 'SoftLayer_Product_Item_Price',
                    'id': price_id}
                prices.append(category)
```

SoftLayer/managers/vs.py
```diff
@@ -874,9 +879,30 @@ def upgrade(self, instance_id, cpus=None, memory=None, nic_speed=None, public=Tr
...
+                for item in vsi_disk.get('billingItem').get('children'):
+                    if item.get('categoryCode').__contains__('guest_disk'):
+                        if disk_number < int("".join(filter(str.isdigit, item.get('categoryCode')))):
+                            disk_number = int("".join(filter(str.isdigit, item.get('categoryCode'))))
+                category = {'categories': [{
+                    'categoryCode': 'guest_disk' + str(disk_number + 1),
+                    'complexType': "SoftLayer_Product_Item_Category"
+                }], 'complexType': 'SoftLayer_Product_Item_Price'}
+                prices.append(category)
+                prices[0]['id'] = price_id
```
allmightyspiff, line 901: You should set the id in the category object. Since its not certain that prices[0] will be the disk price.

SoftLayer/managers/vs.py
```diff
@@ -874,9 +879,30 @@ def upgrade(self, instance_id, cpus=None, memory=None, nic_speed=None, public=Tr
...
+            if disk is not None:
+                category = {'categories': [{
+                    'categoryCode': 'guest_disk' + str(disk.get('number')),
+                    'complexType': "SoftLayer_Product_Item_Category"
+                }], 'complexType': 'SoftLayer_Product_Item_Price'}
+                prices.append(category)
+                prices[0]['id'] = price_id
+            elif add:
+                vsi_disk = self.get_instance(instance_id)
+                disk_number = 0
```
allmightyspiff, line 891: I think you can just set this to `https://sldn.softlayer.com/reference/datatypes/SoftLayer_Virtual_Guest/#blockdevicecount` that way you can make a simple / quick API call instead of doing this nested loop.

SoftLayer/managers/vs.py
```diff
@@ -818,7 +819,8 @@ def capture(self, instance_id, name, additional_disks=False, notes=None):
         return self.guest.createArchiveTransaction(
             name, disks_to_capture, notes, id=instance_id)
 
-    def upgrade(self, instance_id, cpus=None, memory=None, nic_speed=None, public=True, preset=None):
+    def upgrade(self, instance_id, cpus=None, memory=None, nic_speed=None, public=True, preset=None,
```
allmightyspiff, line 822: I dont like having `add` and `disk` here. My suggestion is to have `disks` be an array of disk objects to either upgrade or add, and you can just loop through them to add the prices.

```
disks = [
{'capacity':100, 'number':1}, # upgrade existing disk 1
{'capacity':200, 'number'':-1} # add new disk if the number is -1
]
```

then just remove the `add` parameter.

SoftLayer/CLI/virt/upgrade.py
```diff
@@ -20,15 +22,17 @@
               help="CPU core will be on a dedicated host server.")
 @click.option('--memory', type=virt.MEM_TYPE, help="Memory in megabytes")
 @click.option('--network', type=click.INT, help="Network port speed in Mbps")
+@click.option('--add', type=click.INT, required=False, help="add Hard disk in GB")
```
allmightyspiff, line 25: Change this to `--add-disk` and have it use multiple so they can add more than 1 disk.

SoftLayer/CLI/virt/upgrade.py
```diff
@@ -20,15 +22,17 @@
...
+@click.option('--disk', nargs=1, help="update the number and capacity in GB Hard disk, E.G {'number':2,'capacity':100}")
```
allmightyspiff, line 26: Have this be a Tuple instead of a json string. https://click.palletsprojects.com/en/7.x/options/#tuples-as-multi-value-options. And change the name to --resize-disk to make it clear what it does.

@click.option('--resize-disk', type=(int, int), help="Update disk number to size in GB. --resize-disk 1 250 ")

allmightyspiff:
## TESTING

- :heavy_check_mark: Adding disk
- :heavy_check_mark: Upgrading disk 
- :heavy_check_mark: Adding AND Upgrading disk
- ❌ Handle bad user input

If we get the following error back from the API, `SoftLayerAPIError(500): Price does not have an id.`, I think it would be a good idea to handle that and print out something like "Unable to Upgrade disk `Disk Number Here` to `However many GB`

Also, might be useful to check if the disk they want to upgrade even exists in the first place.

```
./slcli -vvv  vs upgrade --resize-disk 2 200  105826834

curl -u $SL_USER:$SL_APIKEY -X POST -H "Accept: */*" -H "Accept-Encoding: gzip, deflate, compress" -d '{"parameters": [{"complexType": "SoftLayer_Container_Product_Order_Virtual_Guest_Upgrade", "properties": [{"name": "MAINTENANCE_WINDOW", "value": "2020-08-10 19:04:49+0000"}], "virtualGuests": [{"id": 105826834}], "prices": [{"categories": [{"categoryCode": "guest_disk200", "complexType": "SoftLayer_Product_Item_Category"}], "complexType": "SoftLayer_Product_Item_Price", "id": null}]}]}' 'https://api.softlayer.com/rest/v3.1/SoftLayer_Product_Order/placeOrder.json'
SoftLayerAPIError(500): Price does not have an id.
```

Issue #1299, opened 2020-07-14
Ordering price information improvements


allmightyspiff:
What I'd like is a way to display price information in the following commands, possibly with a switch like `--prices`, as adding that information might get the output very cluttered.

- slcli hw create-options
- slcli vs create-optoins
- slcli order item-list

Things to take into implement:
- Hourly and Monthly prices
- Capacity restricted prices
- Possibly let user pick a location to display prices in?

PR #1319, opened 2020-08-11
Ordering price information improvements.

FernandoOjeda:
Ordering price information improvements https://github.com/softlayer/softlayer-python/issues/1299.

Implemented:
  
    slcli hw create-options
    slcli order item-list

Things to take into implement:

    Hourly and Monthly prices
    Capacity restricted prices
    User can pick a location to display prices in

### DIFF
--- a/SoftLayer/CLI/hardware/create_options.py
+++ b/SoftLayer/CLI/hardware/create_options.py
@@ -4,17 +4,17 @@
 import click
 
 from SoftLayer.CLI import environment
-from SoftLayer.CLI import exceptions
 from SoftLayer.CLI import formatting
 from SoftLayer.managers import hardware
 
 
 @click.command()
-@click.option('--prices', '-p', default=False, help='Filter Item Prices, prices(DEFAULT False)')
-@click.option('--location', '-l', type=click.STRING, help='To filter the item prices by location, enter the Location '
-                                                          'keyName e.g. AMSTERDAM02')
+@click.argument('location', required=False)
+@click.option('--prices', '-p', is_flag=True, help='Use --prices to list the server item prices, and '
+                                                   'to list the Item Prices by location, add it to the '
+                                                   '--prices option, e.g. --prices AMSTERDAM02')
 @environment.pass_env
-def cli(env, prices, location):
+def cli(env, prices, location=None):
     """Server order options for a given chassis."""
 
     hardware_manager = hardware.HardwareManager(env.client)
@@ -30,18 +30,14 @@ def cli(env, prices, location):
         dc_table.add_row([location_info['name'], location_info['key']])
     tables.append(dc_table)
 
-    if location and prices:
-        raise exceptions.CLIAbort("Please select --prices or --location to get the prices, not both")
-
     if prices:
         _preset_prices_table(options['sizes'], tables)
         _os_prices_table(options['operating_systems'], tables)
         _port_speed_prices_table(options['port_speeds'], tables)
         _extras_prices_table(options['extras'], tables)
-    elif location:
-        _preset_prices_table(options['sizes'], tables)
-        location_prices = hardware_manager.get_hardware_item_prices(location)
-        _location_item_prices(location_prices, tables)
+        if location:
+            location_prices = hardware_manager.get_hardware_item_prices(location)
+            _location_item_prices(location_prices, tables)
     else:
         # Presets
         preset_table = formatting.Table(['Size', 'Value'], title="Sizes")
@@ -98,20 +94,20 @@ def _os_prices_table(operating_systems, tables):
     :param [] operating_systems: List of Hardware Server operating systems.
     :param tables: Table formatting.
     """
-    os_prices_table = formatting.Table(['OS Key', 'Hourly', 'Monthly', 'capacityRestrictionMaximum',
-                                        'capacityRestrictionMinimum', 'capacityRestrictionType'],
+    os_prices_table = formatting.Table(['OS Key', 'Hourly', 'Monthly', 'Restriction'],
                                        title="Operating Systems Prices")
     os_prices_table.sortby = 'OS Key'
     os_prices_table.align = 'l'
     for operating_system in operating_systems:
         for price in operating_system['prices']:
+            cr_max = _get_price_data(price, 'capacityRestrictionMaximum')
+            cr_min = _get_price_data(price, 'capacityRestrictionMinimum')
+            cr_type = _get_price_data(price, 'capacityRestrictionType')
             os_prices_table.add_row(
                 [operating_system['key'],
                  _get_price_data(price, 'hourlyRecurringFee'),
                  _get_price_data(price, 'recurringFee'),
-                 _get_price_data(price, 'capacityRestrictionMaximum'),
-                 _get_price_data(price, 'capacityRestrictionMinimum'),
-                 _get_price_data(price, 'capacityRestrictionType')])
+                 "%s - %s %s" % (cr_min, cr_max, cr_type)])
     tables.append(os_prices_table)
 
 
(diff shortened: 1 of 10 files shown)


FernandoOjeda:
I think the `slcli vs create-optoins` has to be improve the same that `slcli hw create-options` to add the prices information:

image

Or maybe I can add a new request to shows the item prices in another table and not to add then to the existing table.

allmightyspiff:
Yeah I agree about `slcli vs|hw create-options`, I'll have to think about it a bit and decide how I want the output to look while I review your current changes.

SoftLayer/CLI/hardware/create_options.py
```diff
@@ -4,13 +4,17 @@
 import click
 
 from SoftLayer.CLI import environment
+from SoftLayer.CLI import exceptions
 from SoftLayer.CLI import formatting
 from SoftLayer.managers import hardware
 
 
 @click.command()
+@click.option('--prices', '-p', default=False, help='Filter Item Prices, prices(DEFAULT False)')
```
allmightyspiff, line 13: lets combine prices and location.
so if I do `--prices` I get all default price output. if I do `--prices mex01` I get mex01 prices for example. Might require using a callback function for `--prices` with no value after... not sure. 

Also, when using a location price, the default price should still show up if there isn't a location based one.

SoftLayer/CLI/hardware/create_options.py
```diff
@@ -22,41 +26,171 @@ def cli(env):
...
+    tables.append(preset_prices_table)
+
+
+def _os_prices_table(operating_systems, tables):
+    """Shows Server Operating Systems prices cost and capacity restriction.
+
+    :param [] operating_systems: List of Hardware Server operating systems.
+    :param tables: Table formatting.
+    """
+    os_prices_table = formatting.Table(['OS Key', 'Hourly', 'Monthly', 'capacityRestrictionMaximum',
```
allmightyspiff, line 101: To make this easier to read, lets make the capacity restriction part a bit more condensed for all tables with this field

```
:................................................................................:
: Operating Systems Prices                                                       :
:.............................................:........:.........:...............:
: OS Key                                      : Hourly : Monthly : Restriction   : 
:.............................................:........:.........:...............:
: OS_CENTOS_6_X_32_BIT                        : 0      : 0       : None          :
: OS_WINDOWS_2008_FULL_WEB_64_BIT_R2_SP1      : -      : 11      : 1-1 PROCESSOR :
: OS_WINDOWS_2008_FULL_WEB_64_BIT_R2_SP1      : -      : 22      : 2-2 PROCESSOR :
: OS_WINDOWS_2012_FULL_STD_64_BIT             : .056   : 36      : 2-4 CORE      :
: OS_WINDOWS_2012_FULL_STD_64_BIT             : .112   : 72      : 6-16 CORE     :
:.............................................:........:.........:...............:
```

SoftLayer/CLI/hardware/create_options.py
```diff
@@ -22,41 +26,171 @@ def cli(env):
...
+    """Shows Server Preset options prices.
+
+    :param [] sizes: List of Hardware Server sizes.
+    :param tables: Table formatting.
+    """
+    preset_prices_table = formatting.Table(['Size', 'Value', 'Hourly', 'Monthly'], title="Sizes Prices")
+    preset_prices_table.sortby = 'Value'
+    preset_prices_table.align = 'l'
+    for size in sizes:
+        preset_prices_table.add_row([size['name'], size['key'], size['hourlyRecurringFee'], size['recurringFee']])
```
allmightyspiff, line 91: recurringFees should have as many decimals in their output as from the API output. Which I think is 4 decimal places.

SoftLayer/CLI/order/item_list.py
```diff
@@ -3,19 +3,25 @@
...
 COLUMNS = ['category', 'keyName', 'description', 'priceId']
+COLUMNS_ITEM_PRICES = ['keyName', 'priceId', 'Hourly', 'Monthly', 'CRMax', 'CRMim', 'CRType']
+COLUMNS_ITEM_PRICES_LOCATION = ['keyName', 'priceId', 'Hourly', 'Monthly', 'CRMax', 'CRMim', 'CRType']
 
 
 @click.command()
 @click.argument('package_keyname')
 @click.option('--keyword', help="A word (or string) used to filter item names.")
 @click.option('--category', help="Category code to filter items by")
+@click.option('--prices', '-p', default=False, help='Filter Item Prices, prices(DEFAULT False)')
```
allmightyspiff, line 20: Same comment as `sl hw create-options` about combining these.

SoftLayer/CLI/order/item_list.py
```diff
@@ -47,9 +55,22 @@ def cli(env, package_keyname, keyword, category):
     sorted_items = sort_items(items)
 
     categories = sorted_items.keys()
-    for catname in sorted(categories):
-        for item in sorted_items[catname]:
-            table.add_row([catname, item['keyName'], item['description'], get_price(item)])
+    if prices:
+        table = formatting.Table(COLUMNS_ITEM_PRICES, title="CRMax = CapacityRestrictionMaximum, "
```
allmightyspiff, line 59: As with `create-options` combine the capacity columns in to a single column called `Restrictions` with the format `{Min}-{Max} {TYPE}`

allmightyspiff:
Thanks, this is a big improvement I think. We'll tackle the `vs create-options` in another issue.

Big things to change:

1. `--prices` should take either 0 arguments, meaning default prices. Or 1 argument, meaning location specific prices
2. Capacity restrictions should be combined into a single column to make it easier to read
3. When using location specific pricing, default pricing should still be shown if there is no location pricing. Hopefully that isn't too complicated to add.

SoftLayer/CLI/hardware/create_options.py
FernandoOjeda: Done.

SoftLayer/CLI/hardware/create_options.py
FernandoOjeda: Done.

SoftLayer/CLI/hardware/create_options.py
FernandoOjeda: recurringFees seems to be all integer values, I did not see any recurringFees with decimal values by api output and printing this option by the code.
Please could you check for me this.

SoftLayer/CLI/order/item_list.py
FernandoOjeda: Done.

SoftLayer/CLI/order/item_list.py
FernandoOjeda: Done.

SoftLayer/CLI/hardware/create_options.py
FernandoOjeda: Do you mean this?
image
image

FernandoOjeda: Done.

allmightyspiff:
## TESTS
✔️ `hw create-options --prices`
✔️ `hw create-options --prices dal13`
✔️ `hw create-options --prices tok02`
✔️ `order item-list --prices AMSTERDAM02  SINGLE_E31270_V6`
❌ `order item-list SINGLE_E31270_V6  --prices dal13`
I get the error `Package dal13 does not exist`, not quite sure from looking at the code why click isn't processing that field, maybe because you have `type=click.UNPROCESSED`?



## Future Issues
Once this is merged I'll make issues for these other improvements. This change set is already getting kinda large and complex.

- `hw create-options --prices tok02` lists a few 0$ servers. I think that is because those are not available in tok02, as the others are priced correctly. These should be removed, or labeled as `not available` some way. We can solve that in a different issue though, since this one is already a big change.


```
$ ./slcli --format=table hardware create-options --prices tok02
:....................:
:    Datacenters     :
:............:.......:
: Datacenter : Value :
:............:.......:
: Tokyo 2    : tok02 :
:............:.......:
:.....................................................................................:
:                                     Sizes Prices                                    :
:...............................:.................................:........:..........:
: Size                          : Value                           : Hourly : Monthly  :
:...............................:.................................:........:..........:
: 1U 1270v3 32GB 2x2TB No-RAID  : 1U_1270_V3_2X2TB_NORAID         : 0.0000 : 0.0000   :
: 1U 1270v6 32GB 2x2TB No-RAID  : 1U_1270_V6_2X2TB_NORAID         : 0.0000 : 0.0000   :
: 1U 2174s x 64GB, 2x4TB, RAID1 : 1U_2174S_64GB_2X4TB_RAID_1      : 0.0770 : 57.0000  :
: 1U 4210s, 384GB, 2x4TB, RAID1 : 1U_4210S_384GB_2X4TB_RAID_1     : 0.0600 : 39.5500  :
: 1U 6248 384GB 1x3.8 SED JBOD  : 1U_6248S_384GB_1X3_8_SED_JBOD   : 0.4260 : 283.0000 :
: 1U 6248s, 384GB, 2x4TB, RAID1 : 1U_6248S_384GB_2X4TB_RAID_1     : 0.0770 : 57.0000  :
: 1U 6248 384 4x3.8 SED RAID-10 : 1U_6248S_384GB_4X_3_8TB_RAID_10 : 0.7400 : 492.0000 :
:...............................:.................................:........:..........:
```

- break out `order item-list`  by category, like `hw create-options`. Right now they are sorted by category, but I really like how `hw create-options` looks now. And Right Align the keyname/description fields to be easier to read.
- I'd like `item-list` to behave like `create-options` with the datacenter name as well, instead of the keyname. I think most customers expect `dal13` form, not the longer keyname form.
- I'd like for the location specific prices to just be in the first table (replacing their generic item when applicable).

FernandoOjeda:
It is a good idea to open another issue for hw create-options --prices tok02 to avoid to lists the 0$ data.
And it would be good to open another issue to refactor the order item-list the same than hardware create-options.

PR #1333, opened 2020-08-25
Set notes for network storage

ATGE:


allmightyspiff:
- ✔️ `slcli file volume-list`
- ✔️ `slcli file volume-detail`
- ✔️ `slcli file volume-set-note`
- ✔️ `slcli block volume-list`
- ✔️ `slcli block volume-detail`
- ✔️ `slcli block volume-set-note`

Issue #999, opened 2018-07-12
unexpected error on dns record-list


fabiograssofy:
Hello,
I receive an error when using the dns record-list command. Some zones are working fine but with other not. It seems that the errors is related to zone with number of records  >80/100.

### Expected Behavior
Get the records list

### Actual Behavior
Receive this error:

```
# slcli  dns  record-list domain.com
An unexpected error has occured:
Traceback (most recent call last):
  File "/usr/lib64/python2.7/site-packages/SoftLayer/CLI/core.py", line 182, in main
    cli.main(**kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python2.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/usr/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib64/python2.7/site-packages/SoftLayer/CLI/dns/record_list.py", line 43, in cli
    the_record['host'],
KeyError: 'host'

Feel free to report this error as it is likely a bug:
    https://github.com/softlayer/softlayer-python/issues
The following snippet should be able to reproduce the error
```

### Environment Information
Operating System: tested on CentOS Linux 7.4.1708 and Ubuntu 14.04.5 LTS
softlayer-python version (`slcli --version`): tested with 5.5.0 and 5.3.2

allmightyspiff:
would you mind sharing one of the domains that generates this error? If you don't want to post the domain here, just the zone id should be enough for me to figure out what is causing this error.

fabiograssofy:
Hello @allmightyspiff, I prefer to not share details in a public site.

Can I send you an email (an @softlayer.com address would be better)?

[...]

PR #1335, opened 2020-08-27
Some improvements to the dns commands

allmightyspiff:
- dns zone-list: added resourceRecordCount, added automatic pagination for large zones
- dns record-list: fixed an issue where a record (like SRV types) that don't have a `host` would cause the command to fail


allmightyspiff:
Fixes #999

PR #1337, opened 2020-09-01
refactor vsi create-option

caberos:
https://github.com/softlayer/softlayer-python/issues/1334


SoftLayer/CLI/virt/create_options.py
```diff
@@ -6,213 +6,92 @@
 import SoftLayer
 from SoftLayer.CLI import environment
 from SoftLayer.CLI import formatting
-from SoftLayer import utils
 
 
 @click.command(short_help="Get options to use for creating virtual servers.")
+@click.option('--vsi-type', required=False, type=click.Choice(['TRANSIENT', 'SUSPEND']),
```
allmightyspiff, line 12: Add `PUBLIC` as a choice, and make it the default here, instead of in the cli command. Add `show_default=True` to the option.

SoftLayer/managers/vs.py
```diff
@@ -249,7 +249,7 @@ def get_instance(self, instance_id, **kwargs):
         return self.guest.getObject(id=instance_id, **kwargs)
 
     @retry(logger=LOGGER)
-    def get_create_options(self):
+    def get_create_options(self, vsi_type):
```
allmightyspiff, line 252: set a default for vsi_type. `vsi_type="PUBLIC_CLOUD_SERVER"`

SoftLayer/CLI/virt/create_options.py
```diff
@@ -6,213 +6,92 @@
...
+              help="Billing rate")
```
allmightyspiff, line 13: help should be "Display options for a specific virtual server packages"

SoftLayer/managers/vs.py
```diff
@@ -260,7 +260,94 @@ def get_create_options(self):
...
+
+        return {
+            'locations': locations,
+            'ram': ram,
+            'database': database,
+            'operating_systems': operating_systems,
+            'guest_core': guest_core,
+            'port_speed': port_speeds,
+            'guest_disk': local_disk,
+            'flavors': self.guest.getCreateObjectOptions()['flavors']
```
allmightyspiff, line 335: I'd like to avoid using `getCreateObjectOptions` because that API call takes too long. ` SoftLayer_Virtual_Guest::getCreateObjectOptions (23.201807s)`. Flavors are in the packages getActivePresets

SoftLayer/managers/vs.py
```diff
@@ -260,7 +260,94 @@ def get_create_options(self):
...
+        """Get the package related to simple hardware ordering."""
+        mask = '''
+            items[
+                description, keyName, capacity,
+                attributes[id,attributeTypeKeyName],
+                itemCategory[ id, categoryCode],
+                softwareDescription[id,referenceCode,longDescription],prices],
+                 regions[location[location[priceGroups]]]
+            '''
+        package = self.ordering_manager.get_package_by_key(package_keyname, mask=mask)
```
allmightyspiff, line 349: Lets us self.ordering_manager.get_package_by_key to just get the package Id.


```
package_id = self.ordering_manager.get_package_by_key(package_keyname, mask="mask[id]")
package = self.client.call('Product_Package', 'getObject', id=package_id, mask=mask)
```

SoftLayer/managers/vs.py
```diff
@@ -260,7 +260,94 @@ def get_create_options(self):
...
+            'guest_core': guest_core,
+            'port_speed': port_speeds,
+            'guest_disk': local_disk,
+            'flavors': self.guest.getCreateObjectOptions()['flavors']
+        }
+
+    @retry(logger=LOGGER)
+    def _get_package(self, package_keyname):
+        """Get the package related to simple hardware ordering."""
+        mask = '''
```
allmightyspiff, line 341: Add `activePresets` to this mask, and use it for the flavors.

allmightyspiff (changes requested):
Thanks, changes look mostly great, just a few minor changes to make this command perform better.

If we can get this command to stop using `Virtual_Guest::getCreateObjectOptions` it will actually respond quickly, which will be nice.

allmightyspiff (approved):

Issue #1341, opened 2020-09-08
slcli account orders


allmightyspiff:
`slcli account orders`

should do the following

- list orders and some information about them
- use `iter=True` for paging through them (default to only 50 results)
- support option to filter by date
- support option to show X number of results
- support option to filter by user who placed the order
- table view should be similar to https://cloud.ibm.com/billing/orders

PR #1349, opened 2020-09-22
slcli account orders

caberos:
https://github.com/softlayer/softlayer-python/issues/1341

### DIFF
--- a/SoftLayer/CLI/account/orders.py
+++ b/SoftLayer/CLI/account/orders.py
@@ -11,7 +11,7 @@ from SoftLayer import utils
 
 @click.command()
 @click.option('--limit', '-l',
-              help='How many results to get in one api call, default is 100',
+              help='How many results to get in one api call',
               default=100,
               show_default=True)
 @environment.pass_env
@@ -22,7 +22,6 @@ def cli(env, limit):
 
     order_table = formatting.Table(['Id', 'State', 'User', 'Date', 'Amount', 'Item'],
                                    title="orders")
-    order_table.sortby = 'Amount'
     order_table.align = 'l'
 
     for order in orders:
--- a/SoftLayer/CLI/account/orders.py
+++ b/SoftLayer/CLI/account/orders.py
@@ -1,0 +1,38 @@
+"""Order list account"""
+# :license: MIT, see LICENSE for more details.
+
+import click
+
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+from SoftLayer.managers.account import AccountManager as AccountManager
+from SoftLayer import utils
+
+
+@click.command()
+@click.option('--limit', '-l',
+              help='How many results to get in one api call, default is 100',
+              default=100,
+              show_default=True)
+@environment.pass_env
+def cli(env, limit):
+    """Order list account."""
+    manager = AccountManager(env.client)
+    orders = manager.get_account_all_billing_orders(limit)
+
+    table = []
+    order_table = formatting.Table(['id', 'State', 'user', 'PurchaseDate', 'orderTotalAmount', 'Items'],
+                                   title="orders")
+    order_table.sortby = 'orderTotalAmount'
+    order_table.align = 'l'
+
+    for order in orders:
+        items = []
+        for item in order['items']:
+            items.append(item['description'])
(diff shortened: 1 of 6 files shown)


SoftLayer/CLI/account/orders.py
```diff
@@ -0,0 +1,38 @@
...
+
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+from SoftLayer.managers.account import AccountManager as AccountManager
+from SoftLayer import utils
+
+
+@click.command()
+@click.option('--limit', '-l',
+              help='How many results to get in one api call, default is 100',
```
allmightyspiff, line 14: Dont need `default is 100` in this help text, as the `show_default=True` adds that.

SoftLayer/CLI/account/orders.py
```diff
@@ -0,0 +1,38 @@
...
+              default=100,
+              show_default=True)
+@environment.pass_env
+def cli(env, limit):
+    """Order list account."""
+    manager = AccountManager(env.client)
+    orders = manager.get_account_all_billing_orders(limit)
+
+    table = []
+    order_table = formatting.Table(['id', 'State', 'user', 'PurchaseDate', 'orderTotalAmount', 'Items'],
```
allmightyspiff, line 24: Change column name to 

`order_table = formatting.Table(['Id', 'State', 'User', 'Date', 'Amount', 'Item'],`

SoftLayer/CLI/account/orders.py
```diff
@@ -0,0 +1,38 @@
...
+                                   title="orders")
+    order_table.sortby = 'orderTotalAmount'
```
allmightyspiff, line 26: just keep the default sort from the API, which should be by date, DESC. 

Might be interesting to add an option to sort by amount, and have the API get the most expensive orders with a filter, but that would be another issue. For now the default sort by date is what I would like.

SoftLayer/managers/account.py
```diff
@@ -233,3 +233,28 @@ def cancel_item(self, identifier, reason="No longer needed", note=None):
             note = "Cancelled by {} with the SLCLI".format(user.get('username'))
 
         return self.client.call('Billing_Item', 'cancelItem', False, True, reason, note, id=identifier)
+
+    def get_account_all_billing_orders(self, limit, mask=None):
```
allmightyspiff, line 237: set a default limit in the function declaration.

SoftLayer/CLI/account/orders.py
```diff
@@ -0,0 +1,38 @@
...
+    for order in orders:
+        items = []
+        for item in order['items']:
+            items.append(item['description'])
+        create_date = utils.clean_time(order['createDate'], in_format='%Y-%m-%d', out_format='%Y-%m-%d')
+
+        order_table.add_row([order['id'], order['status'], order['userRecord']['username'], create_date,
+                             order['orderTotalAmount'], utils.trim_to(' '.join(map(str, items)), 50)])
+    table.append(order_table)
+    env.fout(formatting.listing(table, separator='\n'))
```
allmightyspiff, line 38: I don't think the `table` variable is needed. You can just do `env.fout(order_table)` for the same effect

allmightyspiff (changes requested):
A few minor comments, otherwise works great, thanks.

### Tests
✔️ `slcli account orders`
✔️ `slcli account orders --limit 10`

SoftLayer/CLI/account/orders.py
```diff
@@ -0,0 +1,36 @@
...
+              show_default=True)
+@environment.pass_env
+def cli(env, limit):
+    """Order list account."""
+    manager = AccountManager(env.client)
+    orders = manager.get_account_all_billing_orders(limit)
+
+    order_table = formatting.Table(['Id', 'State', 'User', 'Date', 'Amount', 'Item'],
+                                   title="orders")
+    order_table.sortby = 'Amount'
```
allmightyspiff, line 25: Remove this line.

The results from the API are already sorted by date, which is what customers will expect

SoftLayer/CLI/account/orders.py
allmightyspiff: set the help message to `help='How many results to get in one api call'`

allmightyspiff (changes requested):

allmightyspiff (approved):

Issue #1339, opened 2020-09-08
Support `file|block volume-list --order`


allmightyspiff:
ibmcloud cli has the following option we should also add.

```
Order 64570100 was placed.
 > Storage as a Service
 > Block Storage
 > 25 GBs
 > 200 IOPS
 > 20 GB (Snapshot Space)

You may run 'C:\Program Files\IBM\Cloud\bin\ibmcloud.exe sl block volume-list --order 64570100' to find this block volume after it is ready.

 ibmcloud.exe sl block volume-list --order 64570100
2020/09/08 17:34:13 [DEBUG] Request URL:  GET https://api.softlayer.com/rest/v3.1/SoftLayer_Account/getIscsiNetworkStorage.json?objectFilter=%7B%22iscsiNetworkStorage%22%3A%7B%22billingItem%22%3A%7B%22orderItem%22%3A%7B%22order%22%3A%7B%22id%22%3A%7B%22operation%22%3A64570100%7D%7D%7D%7D%2C%22storageType%22%3A%7B%22keyName%22%3A%7B%22operation%22%3A%22%2A%3D+BLOCK_STORAGE%22%7D%7D%7D%7D&objectMask=mask%5Bid%2Cusername%2CserviceResource.datacenter.name%2CstorageType.keyName%2CcapacityGb%2CbytesUsed%2ClunId%5D
2020/09/08 17:34:13 [DEBUG] Parameters:
2020/09/08 17:34:14 [DEBUG] Status Code:  200
2020/09/08 17:34:14 [DEBUG] Response:  [{"capacityGb":25,"id":168669398,"username":"SL02SL307608-38","lunId":"1","serviceResource":{"id":7779,"name":"Storage Type 02 Block Aggregate stbf-dal0903a","datacenter":{"name":"dal09"},"type":{"type":"NETAPP_STOR_AGGR"}},"storageType":{"keyName":"PERFORMANCE_BLOCK_STORAGE"}}]
id          username          datacenter   storage_type                capacity_gb   bytes_used   lunId
168669398   SL02SL307608-38   dal09        performance_block_storage   25            -            1
```

PR #1350, opened 2020-09-22
Add to block and file storage volume list order.

FernandoOjeda:
Add to block and file storage volume list order https://github.com/softlayer/softlayer-python/issues/1339.

### DIFF
--- a/SoftLayer/managers/block.py
+++ b/SoftLayer/managers/block.py
@@ -25,13 +25,13 @@ class BlockStorageManager(StorageManager):
         """
         return self.get_volume_count_limits()
 
-    def list_block_volumes(self, datacenter=None, username=None, order=None, storage_type=None, **kwargs):
+    def list_block_volumes(self, datacenter=None, username=None, storage_type=None, order=None, **kwargs):
         """Returns a list of block volumes.
 
-        :param order: Volume order id.
         :param datacenter: Datacenter short name (e.g.: dal09)
         :param username: Name of volume.
         :param storage_type: Type of volume: Endurance or Performance
+        :param order: Volume order id.
         :param kwargs:
         :return: Returns a list of block volumes.
         """
--- a/SoftLayer/managers/block.py
+++ b/SoftLayer/managers/block.py
@@ -25,9 +25,10 @@ class BlockStorageManager(StorageManager):
         """
         return self.get_volume_count_limits()
 
-    def list_block_volumes(self, datacenter=None, username=None, storage_type=None, **kwargs):
+    def list_block_volumes(self, datacenter=None, username=None, order=None, storage_type=None, **kwargs):
         """Returns a list of block volumes.
 
+        :param order: Volume order id.
         :param datacenter: Datacenter short name (e.g.: dal09)
         :param username: Name of volume.
         :param storage_type: Type of volume: Endurance or Performance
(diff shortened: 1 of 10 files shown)


SoftLayer/managers/block.py
```diff
@@ -25,9 +25,10 @@ def list_block_volume_limit(self):
         """
         return self.get_volume_count_limits()
 
-    def list_block_volumes(self, datacenter=None, username=None, storage_type=None, **kwargs):
+    def list_block_volumes(self, datacenter=None, username=None, order=None, storage_type=None, **kwargs):
```
allmightyspiff, line 28: the new parameter `order` should go after `storage_type`, not before. This way anyone passing in arguments without the names wont encounter an error.

SoftLayer/managers/file.py
```diff
@@ -22,9 +22,10 @@ def list_file_volume_limit(self):
         """
         return self.get_volume_count_limits()
 
-    def list_file_volumes(self, datacenter=None, username=None, storage_type=None, **kwargs):
+    def list_file_volumes(self, datacenter=None, username=None, order=None, storage_type=None, **kwargs):
```
allmightyspiff, line 25: the new parameter order should go after storage_type, not before. This way anyone passing in arguments without the names wont encounter an error.

tests/CLI/modules/block_tests.py
```diff
@@ -141,6 +141,27 @@ def test_volume_list(self):
             }],
             json.loads(result.output))
 
+    def test_volume_list_order(self):
+        result = self.run_command(['block', 'volume-list', '--order=1234567'])
+
+        self.assert_no_fail(result)
+        self.assertEqual([
```
allmightyspiff, line 148: I dont like tests like this, they break too easily if the fixture is changed. 

instead just do 

```python
json_result = json.loads(result.output)
self.assertEqual(json_result['id'], 100)
```

or something. If a few of the fields you expect to be there are correct, likely they rest are as well.

tests/CLI/modules/file_tests.py
```diff
@@ -55,6 +55,26 @@ def test_volume_list(self):
             }],
             json.loads(result.output))
 
+    def test_volume_list_order(self):
+        result = self.run_command(['file', 'volume-list', '--order=1234567'])
+
+        self.assert_no_fail(result)
+        self.assertEqual([
```
allmightyspiff, line 62: I dont like tests like this, they break too easily if the fixture is changed.

instead just do

json_result = json.loads(result.output)
self.assertEqual(json_result['id'], 100)
or something. If a few of the fields you expect to be there are correct, likely they rest are as well.

allmightyspiff (changes requested):
One thing missing is i'd like the output of a storage order to mention that `slcli block volume-list -o <orderid>` can be used to look up this order, like the `ibmcloud` command has.

Thanks.

SoftLayer/managers/block.py
FernandoOjeda: Done.

SoftLayer/managers/file.py
FernandoOjeda: Done.

tests/CLI/modules/block_tests.py
FernandoOjeda: Done.

tests/CLI/modules/file_tests.py
FernandoOjeda: Done.

allmightyspiff (approved):

allmightyspiff:
## TESTING
- ✔️  Command shown when ordering a volume to check order status
- ✔️ Searching by order id works

Issue #1345, opened 2020-09-17
hw create-options --prices tok02 lists a few 0$ servers


allmightyspiff:
Related to #1319 

`hw create-options --prices tok02` lists a few 0$ servers. I think that is because those are not available in tok02, as the others are priced correctly. These should be removed, or labeled as not available some way

allmightyspiff:
```
$ ./slcli --format=table hw create-options --prices tok02
:....................:
:    Datacenters     :
:............:.......:
: Datacenter : Value :
:............:.......:
: Tokyo 2    : tok02 :
:............:.......:
:.....................................................................................:
:                                     Sizes Prices                                    :
:...............................:.................................:........:..........:
: Size                          : Value                           : Hourly : Monthly  :
:...............................:.................................:........:..........:
: 1U 1270v3 32GB 2x2TB No-RAID  : 1U_1270_V3_2X2TB_NORAID         : 0.0000 : 0.0000❌ :
: 1U 1270v6 32GB 2x2TB No-RAID  : 1U_1270_V6_2X2TB_NORAID         : 0.0000 : 0.0000❌ :
: 1U 2174s x 64GB, 2x4TB, RAID1 : 1U_2174S_64GB_2X4TB_RAID_1      : 0.0770 : 57.0000  :
: 1U 4210s, 384GB, 2x4TB, RAID1 : 1U_4210S_384GB_2X4TB_RAID_1     : 0.0600 : 39.5500  :
: 1U 6248 384GB 1x3.8 SED JBOD  : 1U_6248S_384GB_1X3_8_SED_JBOD   : 0.4260 : 283.0000 :
: 1U 6248s, 384GB, 2x4TB, RAID1 : 1U_6248S_384GB_2X4TB_RAID_1     : 0.0770 : 57.0000  :
: 1U 6248 384 4x3.8 SED RAID-10 : 1U_6248S_384GB_4X_3_8TB_RAID_10 : 0.7400 : 492.0000 :
:...............................:.................................:........:..........:
```


Compared to dal13

```
$ ./slcli --format=table hw create-options --prices dal13
:....................:
:    Datacenters     :
:............:.......:
: Datacenter : Value :
:............:.......:
: Dallas 13  : dal13 :
:............:.......:
:......................................................................................:
:                                     Sizes Prices                                     :
:...............................:.................................:........:...........:
: Size                          : Value                           : Hourly : Monthly   :
:...............................:.................................:........:...........:
[...]

PR #1353, opened 2020-09-23
remove the zeros(0) in the list.

caberos:
https://github.com/softlayer/softlayer-python/issues/1345


SoftLayer/CLI/hardware/create_options.py
```diff
@@ -81,8 +81,9 @@ def _preset_prices_table(sizes, tables):
     preset_prices_table.sortby = 'Value'
     preset_prices_table.align = 'l'
     for size in sizes:
-        preset_prices_table.add_row([size['name'], size['key'], "%.4f" % size['hourlyRecurringFee'],
-                                     "%.4f" % size['recurringFee']])
+        if ("%.4f" % size['hourlyRecurringFee'] != '0.0000') or ("%.4f" % size['recurringFee'] != '0.0000'):
```
allmightyspiff, line 84: You can just check if `if (size['hourlyRecurringFee'] > 0) or (size['recurringFee'] > 0):`

allmightyspiff (changes requested):
You only need to remove 0 cost items from the preset table, There can be 0 cost items in the other tables

allmightyspiff (approved):

Issue #1373, opened 2020-11-05
RIR Contact list


allmightyspiff:
## SLCLI registration contacts
lists out the contacts available. Similar to this page: https://cloud.ibm.com/classic/network/rir/records

![image](https://user-images.githubusercontent.com/7408017/98302271-b69e4500-1f81-11eb-987f-89ff26314bfa.png)

### API 
```
slcli --format=json call-api SoftLayer_Account getSubnetRegistrationDetails --mask="mask[detailType,properties[propertyType]]" --json-filter='{"subnetRegistrationDetails":{"detailTypeId":{"operation":3}}}'

SoftLayer_Account::getSubnetRegistrationDetails(id=None, mask='mask[detailType,properties[propertyType]]', filter='{'subnetRegistrationDetails': {'detailTypeId': {'operation': 3}}}', args=(), limit=None, offset=None))
```

getSubnetRegistrationDetails has a lot of information, but the `detailTypeId == 3` seems to be the `People` type records that this portal page shows. This command should list these out with some of the basic information about them in a table.

| id | FirstName LastName | Label | Email | Address (with city/state/country/postal)|
| --- | --- |--- |--- |--- |
| 12345 | Test Testerson | MY Account | myaccount@email.com | 12345 SomeStreet Dallas TX USA 12345 |

allmightyspiff:
```
$ ./slcli --format=table reg contacts
:.........:....................:...........:......................:............................................:..............:
:    Id   :        Name        :   Label   :        Email         :                  Address                   :    Phone     :
:.........:....................:...........:......................:............................................:..............:
:  51990  :  Boberson Jackson  :    None   :         None         :               4849 Alpha Rd.               : 281.714.3156 :
:         :                    :           :                      :             Dallas TX US 75244             :              :
: 1634037 : Sebastian Chaparro : IBM Cloud : schaparro@es.ibm.com : Avenida Diagonal 571 Barcelona OT ES 08029 :  620881761   :
:.........:....................:...........:......................:............................................:..............:
```

PR #1376, opened 2020-11-12
RIR Contact list

caberos:
https://github.com/softlayer/softlayer-python/issues/1373

### DIFF
--- a/SoftLayer/CLI/registration/contacts.py
+++ b/SoftLayer/CLI/registration/contacts.py
@@ -1,0 +1,40 @@
+"""Get Contact details for a subnet registration."""
+# :license: MIT, see LICENSE for more details.
+
+import click
+
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+
+
+@click.command()
+@environment.pass_env
+def cli(env):
+    """Display the RIR contact information for your account.
+
+    """
+    mgr = SoftLayer.NetworkManager(env.client)
+    result = mgr.get_registration_details()
+    table = formatting.Table(['id', 'name', 'email', 'label', 'address', 'phone'])
+    for contact in result:
+        json_result = {'EMAIL_ADDRESS': '', 'INTERNAL_LABEL': ''}
+        json_result['id'] = contact['id']
+        for property_contact in contact['properties']:
+            if property_contact['propertyType']['keyName'] == 'EMAIL_ADDRESS':
+                json_result['EMAIL_ADDRESS'] = property_contact['value']
+                continue
+            elif property_contact['propertyType']['keyName'] == 'INTERNAL_LABEL':
+                json_result['INTERNAL_LABEL'] = property_contact['value']
+                continue
+            json_result[property_contact['propertyType']['keyName']] = (property_contact['value'])
+
+        table.add_row([json_result['id'],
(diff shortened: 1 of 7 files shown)


SoftLayer/CLI/registration/contacts.py
```diff
@@ -0,0 +1,40 @@
...
+
+    """
+    mgr = SoftLayer.NetworkManager(env.client)
+    result = mgr.get_registration_details()
+    table = formatting.Table(['id', 'name', 'email', 'label', 'address', 'phone'])
+    for contact in result:
+        json_result = {'EMAIL_ADDRESS': '', 'INTERNAL_LABEL': ''}
+        json_result['id'] = contact['id']
+        for property_contact in contact['properties']:
+            if property_contact['propertyType']['keyName'] == 'EMAIL_ADDRESS':
```
allmightyspiff, line 24: what is the point of this if/else block? I dont understand why `json_result[property_contact['propertyType']['keyName']] = (property_contact['value'])` doesn't work for EMAIL/INTERNAL_LABEL

SoftLayer/managers/network.py
```diff
@@ -388,6 +388,17 @@ def get_registrations(self):
 
         return self.account.getSubnetRegistrations(mask=mask)
 
+    def get_registration_details(self):
```
allmightyspiff, line 391: Move this function to the registration manager.
Also allow this function to take in a `mask` parameter, and if it is None, set it to the default you have below.

SoftLayer/CLI/registration/contacts.py
```diff
@@ -0,0 +1,40 @@
+"""Get Contact details for a subnet registration."""
```
allmightyspiff, line 1: should be: `Display the RIR contact information for your account.`

SoftLayer/CLI/registration/contacts.py
```diff
@@ -0,0 +1,40 @@
...
+
+@click.command()
+@environment.pass_env
+def cli(env):
+    """Display the RIR contact information for your account.
+
+    """
+    mgr = SoftLayer.NetworkManager(env.client)
+    result = mgr.get_registration_details()
+    table = formatting.Table(['id', 'name', 'email', 'label', 'address', 'phone'])
```
allmightyspiff, line 19: Capitalize the first letter of each column name

SoftLayer/CLI/registration/contacts.py
```diff
@@ -0,0 +1,40 @@
...
+        for property_contact in contact['properties']:
+            if property_contact['propertyType']['keyName'] == 'EMAIL_ADDRESS':
+                json_result['EMAIL_ADDRESS'] = property_contact['value']
+                continue
+            elif property_contact['propertyType']['keyName'] == 'INTERNAL_LABEL':
+                json_result['INTERNAL_LABEL'] = property_contact['value']
+                continue
+            json_result[property_contact['propertyType']['keyName']] = (property_contact['value'])
+
+        table.add_row([json_result['id'],
```
allmightyspiff, line 32: instead of `json_result['id]` use `json_result.get('id')` for this and all other fields in json_result. Due to the way the registration stuff is designed, I think its certainly possible for  some of these fields to just not exist, which would cause an exception for the user.

allmightyspiff (changes requested):

allmightyspiff (approved):

Issue #1395, opened 2020-12-06
Unable to create reserved-guest


allmightyspiff:
```
allmi@SPF-WIN10 MINGW64 ~/Source/githubio_source (20201204)
$ slcli --format=json  vs capacity create-guest --capacity-id 80068 -H cgallo-reserved-test1 -D cgallo.com -o UBUNTU_LATEST
SoftLayerAPIError(SoftLayer_Exception_Order_Item_Invalid): The price 4 GB (216541) is not valid for package (1035).
(py38)
allmi@SPF-WIN10 MINGW64 ~/Source/githubio_source (20201204)
$ slcli --format=json  vs capacity create-guest --capacity-id 6701 -H cgallo-reserved-test1 -D cgallo.com -o UBUNTU_LATEST
SoftLayerAPIError(SoftLayer_Exception_Order_Item_Invalid): The price 8 GB (216553) is not valid for package (1035).
```

Looks like price look ups might be broken or something on creating reserved capacity guest. Looks to be getting the RAM/CORE price instead of using a preset?

allmightyspiff:
Looking at this a bit more, I think the problem is with the behavior of  https://sldn.softlayer.com/reference/services/SoftLayer_Virtual_Guest/generateOrderTemplate/ as it uses the package `SUSPEND_CLOUD_SERVER`,  but reserved capacity requires package `PUBLIC_CLOUD_SERVER`.

I don't see a way to force `SoftLayer_Virtual_Guest/generateOrderTemplate` to use the PUBLIC_CLOUD_SERVER package ( unless its a monthly order, but we want to support hourly since the portal does as well). I suspect this will require refactoring  `SoftLayer.vs_capacity.create_guest()` to use the ordering manager directly, instead of the `vs_manager.verify_create_instance()` method.

PR #1402, opened 2020-12-18
Unable to create reserved-guest

caberos:
https://github.com/softlayer/softlayer-python/issues/1395

### DIFF
--- a/SoftLayer/managers/vs.py
+++ b/SoftLayer/managers/vs.py
@@ -465,7 +465,7 @@ class VSManager(utils.IdentifierMixin, object):
                                 'FORCE', config, id=instance_id)
 
     def _generate_create_dict(
-            self, cpus=None, memory=None, useHourlyPricing=True,
+            self, cpus=None, memory=None, use_hourly_pricing=True,
             hostname=None, domain=None, local_disk=True,
             datacenter=None, os_code=None, image_id=None,
             dedicated=False, public_vlan=None, private_vlan=None,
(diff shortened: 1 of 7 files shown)


SoftLayer/managers/vs.py
```diff
@@ -465,7 +465,7 @@ def reload_instance(self, instance_id,
                                 'FORCE', config, id=instance_id)
 
     def _generate_create_dict(
-            self, cpus=None, memory=None, hourly=True,
+            self, cpus=None, memory=None, useHourlyPricing=True,
```
allmightyspiff: Don't change names of function arguments if we can help it. Change this back to "hourly" so we don't break anyone's code that happens to be using this function.

SoftLayer/managers/vs.py
```diff
@@ -504,7 +504,7 @@ def _generate_create_dict(
             "hostname": hostname,
             "domain": domain,
             "localDiskFlag": local_disk,
-            "hourlyBillingFlag": hourly,
+            "useHourlyPricing": useHourlyPricing,
```
allmightyspiff: Make sure this change doesn't break hourly VSI orders of any kind.

allmightyspiff (changes requested):

allmightyspiff:
I finally had a moment to get back to this, and I decided that instead of doing all the work needed to get reserved capacity to allow hourly creation... were just going to force monthly until someone asks us to do hourly as well.

I'll fix this in #1454

PR #1403, opened 2020-12-18
Added a unit test for large ints

allmightyspiff:
Making this pull request mostly just to test out the builds.

### DIFF
--- a/SoftLayer/fixtures/SoftLayer_Network_Storage_Hub_Cleversafe_Account.py
+++ b/SoftLayer/fixtures/SoftLayer_Network_Storage_Hub_Cleversafe_Account.py
@@ -37,3 +37,12 @@ getCredentials = [
         }
     }
 ]
+
+getBuckets = [
+    {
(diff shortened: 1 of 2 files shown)


ATGE (approved):
LGTM 🎉

eor1994:
Hi Christopher,

Sorry for the late reply on this thread. 

I have tested the code you have attached and have gotten the same result. Please see output result below:

------------------------
('32bit', 'WindowsPE')
Max Int: 2147483647
BytesUsed: 2666148982056
bytesUsed is a <class 'int'>
2666148982056
Decimal: 2666148982056 Int: 2666148982056
v5.9.2
--------------------------
We created a test in postman. 

So this is a request for a buckets attributes directly using the API over a REST connection. As you can see the Bytes used are being returned as a negative number. This indicates it is the API that is wrong, not the client.

postman_api

Please advise on going forward?

Thank you for your time. 

Regards,
Eoin

PR #1405, opened 2020-12-23
Registration update-all

ATGE:


SoftLayer/managers/registration.py
```diff
@@ -36,6 +36,92 @@ def detail(self, identifier):
...
         return None
 
+    def update_all(self, contact_id, skip_subnets=None):
+        """Updates every subnet to contact
+
+        :param list skip_subnets: A list of subnets ids to skip
+        :param int contact_id: Id of a Account_Regional_Registry_Detail that is a person
+        :return:A list of subnet registrations
+        """
+        skip_subnets = [] if skip_subnets is None else skip_subnets
```
allmightyspiff, line 46: instead of doing this `skip_subnets = [] if skip_subnets is None else skip_subnets`, which is hard to read, just set `skip_subnets = []` in the function declaration.

SoftLayer/managers/registration.py
```diff
@@ -36,6 +36,92 @@ def detail(self, identifier):
...
+        }
+
+    def get_public_subnets(self):
+        """Gets public subnets"""
+
+        mask = 'mask[id,networkIdentifier,activeRegistration,regionalInternetRegistry,cidr]'
+        subnets = self.account.getPublicSubnets(mask=mask)
+        return subnets
+
+    def register_subnet(self, subnet, contact_id):
```
allmightyspiff, line 87: This function is just called `register` on my local branch, can you make sure you don't have two functions doing the same thing in your branch.

allmightyspiff (changes requested):

SoftLayer/managers/registration.py
ATGE: done

SoftLayer/managers/registration.py
ATGE: updated

allmightyspiff (approved):

PR #1407, opened 2021-01-06
Adding disaster recovery failover api method

Rajarajan97:


docs/cli/block.rst
```diff
@@ -142,3 +142,7 @@ Block Commands
 .. click:: SoftLayer.CLI.block.set_note:cli
     :prog: block volume-set-note
     :show-nested:
+
```
allmightyspiff, line 145: If there is any cloud.ibm.com/docs information on this feature, can you link it here?

Or at least elaborate a bit on some of the dangers of this command here.

SoftLayer/CLI/block/replication/disaster_recovery_failover.py
```diff
@@ -0,0 +1,32 @@
...
+from SoftLayer.CLI import formatting
+from SoftLayer.CLI import exceptions
+
+
+@click.command()
+@click.argument('volume-id')
+@click.option('--replicant-id', help="ID of the replicant volume")
+@environment.pass_env
+def cli(env, volume_id, replicant_id):
+    """Failover an inaccessible file volume to its available replicant volume."""
```
allmightyspiff, line 16: Might want to mention the dangers of running this command here. And maybe mention to use `replica-failover` for most cases?

allmightyspiff (changes requested):
Code looks good, only complaint is around some of the documentation bits. Should be a bit more clear that this is a sort of worst case scenario command to use.

allmightyspiff (approved):

allmightyspiff:
I will merge this when  FBLOCK-3769 is released.

FernandoOjeda (approved):
The code looks good.

Issue #1409, opened 2021-01-29
Add pagination to `slcli -v object-storage accounts`


allmightyspiff:
`slcli -v object-storage accounts`
should default to returning 10 accounts, and use `iter=True` to get through the rest. A user reported this command timeing out with only 36 accounts, so breaking up the API calls should help.

PR #1411, opened 2021-02-01
Add pagination to object storage list accounts.

FernandoOjeda:
Add pagination to object storage list accounts https://github.com/softlayer/softlayer-python/issues/1409.

### DIFF
--- a/SoftLayer/CLI/object_storage/list_accounts.py
+++ b/SoftLayer/CLI/object_storage/list_accounts.py
@@ -9,12 +9,16 @@ from SoftLayer.CLI import formatting
 
 
 @click.command()
+@click.option('--limit',
+              type=int,
+              default=10,
+              help="Result limit")
 @environment.pass_env
-def cli(env):
+def cli(env, limit):
     """List object storage accounts."""
 
     mgr = SoftLayer.ObjectStorageManager(env.client)
-    accounts = mgr.list_accounts()
+    accounts = mgr.list_accounts(limit=limit)
     table = formatting.Table(['id', 'name', 'apiType'])
     table.sortby = 'id'
     api_type = None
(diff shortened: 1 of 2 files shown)


ATGE (approved):
Testing
✔️ slcli -v object-storage accounts

LGTM 🎉

SoftLayer/managers/object_storage.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, client):
     def list_accounts(self):
         """Lists your object storage accounts."""
         return self.client.call('Account', 'getHubNetworkStorage',
-                                mask=LIST_ACCOUNTS_MASK)
+                                mask=LIST_ACCOUNTS_MASK, iter=True, limit=10)
```
caberos, line 33: I think the limit should be sent as a method parameter but it should have a default value in case not sent  
 E.g.
          def list_accounts(self, limit=10):

allmightyspiff: Given how slow this api seems to be, having a configurable limit is a good idea.

allmightyspiff (changes requested):
Just add a `--limit NUM` option (with a default of 10) and this is good to go. Thanks.

SoftLayer/managers/object_storage.py
FernandoOjeda: Done.

SoftLayer/CLI/object_storage/list_accounts.py
```diff
@@ -9,12 +9,16 @@
 
 
 @click.command()
+@click.option('--limit',
+              type=int,
+              default=10,
```
caberos, line 14: I think remove this, only should have a default parameter 
`def list_accounts(self, limit=10):`

allmightyspiff: Having limit be a click option is what is needed. That allows users to change the limit when running the command.

allmightyspiff (approved):

Issue #1412, opened 2021-02-01
Create a Virtual Server by backend/frondend router id.


FernandoOjeda:
Add to `slcli vs create` the option to order by backend/frondend router id.

PR #1414, opened 2021-02-05
Add slcli vs create by router data.

FernandoOjeda:
Add slcli vs create by router data https://github.com/softlayer/softlayer-python/issues/1412.

### DIFF
--- a/SoftLayer/managers/vs.py
+++ b/SoftLayer/managers/vs.py
@@ -470,7 +470,6 @@ class VSManager(utils.IdentifierMixin, object):
             datacenter=None, os_code=None, image_id=None,
             dedicated=False, public_vlan=None, private_vlan=None,
             private_subnet=None, public_subnet=None,
-            public_router=None, private_router=None,
             userdata=None, nic_speed=None, disks=None, post_uri=None,
             private=False, ssh_keys=None, public_security_groups=None,
             private_security_groups=None, boot_mode=None, transient=False, **kwargs):
@@ -534,13 +533,14 @@ class VSManager(utils.IdentifierMixin, object):
         if datacenter:
             data["datacenter"] = {"name": datacenter}
 
-        if private_router or public_router:
+        if kwargs.get('private_router') or kwargs.get('public_router'):
             if private_vlan or public_vlan or private_subnet or public_subnet:
                 raise exceptions.SoftLayerError("You have to select network vlan or network vlan with a subnet or "
                                                 "only router, not all options")
             network_components = self._create_network_components(public_vlan, private_vlan,
                                                                  private_subnet, public_subnet,
-                                                                 private_router, public_router)
+                                                                 kwargs.get('private_router'),
+                                                                 kwargs.get('public_router'))
             data.update(network_components)
 
         if private_vlan or public_vlan or private_subnet or public_subnet:
--- a/SoftLayer/managers/vs.py
+++ b/SoftLayer/managers/vs.py
@@ -533,20 +533,7 @@ class VSManager(utils.IdentifierMixin, object):
         if datacenter:
             data["datacenter"] = {"name": datacenter}
 
-        if kwargs.get('private_router') or kwargs.get('public_router'):
-            if private_vlan or public_vlan or private_subnet or public_subnet:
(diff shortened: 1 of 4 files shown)


SoftLayer/managers/vs.py
```diff
@@ -533,10 +533,7 @@ def _generate_create_dict(
         if datacenter:
             data["datacenter"] = {"name": datacenter}
 
-        if private_vlan or public_vlan or private_subnet or public_subnet:
-            network_components = self._create_network_components(public_vlan, private_vlan,
-                                                                 private_subnet, public_subnet)
-            data.update(network_components)
+        self.get_network_components(data, kwargs, private_subnet, private_vlan, public_subnet, public_vlan)
```
allmightyspiff, line 536: I don't like function that modify the data that gets passed into them. I'd prefer this function `get_network_componets` return `data` instead.

So this line would be:
```
network_components = self.get_network_components(kwargs, private_subnet, private_vlan, public_subnet, public_vlan)
data.update(network_components)
```
as that reads better to me.

SoftLayer/managers/vs.py
```diff
@@ -579,9 +576,34 @@ def _generate_create_dict(
 
         return data
 
+    def get_network_components(self, data, kwargs, private_subnet, private_vlan, public_subnet, public_vlan):
+        """Get the network components structure.
+
+        :param data: Array variable to add the network structure.
```
allmightyspiff, line 582: Have this be the `Return` section

SoftLayer/managers/vs.py
```diff
@@ -579,9 +576,34 @@ def _generate_create_dict(
...
+        :param kwargs: Vs item list.
+        :param int private_subnet: Private subnet id.
+        :param int private_vlan: Private vlan id.
+        :param int public_subnet: Public subnet id.
+        :param int public_vlan: Public vlan id.
+        """
+        if kwargs.get('private_router') or kwargs.get('public_router'):
```
allmightyspiff, line 589: For readability, change

```python
if kwargs.get('private_router') or kwargs.get('public_router'):
    if private_vlan or public_vlan or private_subnet or public_subnet:
```
To:

```python
# Throw exception for router option AND vlan/subnet option
if any([kwargs.get('private_router'), kwargs.get('public_router')] and any([private_vlan, public_vlan, private_subnet, public_subnet]):

```

SoftLayer/managers/vs.py
```diff
@@ -579,9 +576,34 @@ def _generate_create_dict(
 
         return data
 
+    def get_network_components(self, data, kwargs, private_subnet, private_vlan, public_subnet, public_vlan):
```
allmightyspiff, line 579: I think you could combine this function with `_create_network_components`,  `kwargs.get()` should rerturn `None` if the option wasn't specified. So really you just need to move the Exception where your checking for vlan/subnets and router to `create_network_components` and save having to make another function.

allmightyspiff (changes requested):
Looks good, but I would recommend combining `get_network_components` and `create_network_components` into one function. Basically just add your check for the exception to `create_network_components` and I think it wil work as you expect.

SoftLayer/managers/vs.py
FernandoOjeda: I did it the same lol, but many unit tests failed. I am going to check it again.

ATGE (approved):
LGTM

SoftLayer/managers/vs.py
FernandoOjeda: Done.

SoftLayer/managers/vs.py
FernandoOjeda: Done.

SoftLayer/managers/vs.py
FernandoOjeda: Done.

SoftLayer/managers/vs.py
FernandoOjeda: Done.

allmightyspiff (approved):
Looks good, thanks

caberos (approved):
Good job

PR #1415, opened 2021-02-05
Add username lookup to slcli  object-storage credential

ATGE:


SoftLayer/managers/object_storage.py
```diff
@@ -26,11 +29,16 @@ class ObjectStorageManager(object):
 
     def __init__(self, client):
         self.client = client
+        self.resolvers = [self._get_id_from_username]
 
-    def list_accounts(self, limit=10):
+    def list_accounts(self, object_mask=None, object_filter=None, limit=10):
```
caberos, line 34: I think the `object_mask `should be  to `LIST_ACCOUNTS_MASK `and not `none`

allmightyspiff: `object_mask = object_mask if object_mask else LIST_ACCOUNTS_MASK` sets the mask to LIST_ACCOUNT_MASK if one wasn't passed in. 

Given how long Object masks tend to be, I don't mind doing the `if none: objectMask = defaultMask` pattern instead of setting the default in the function definition.

allmightyspiff (approved):

FernandoOjeda (approved):
It looks good for me.

caberos (approved):

Issue #1440, opened 2021-03-18
Authorize block and file storage to a virtual server 


FernandoOjeda:
Authorize block and file storage to a virtual server.

PR #1442, opened 2021-03-22
Add authorize block, file and portable storage to a vs.

FernandoOjeda:
Add authorize block, file and portable storage to a vs https://github.com/softlayer/softlayer-python/issues/1440.

### DIFF
--- a/SoftLayer/managers/vs.py
+++ b/SoftLayer/managers/vs.py
@@ -1404,3 +1404,45 @@ class VSManager(utils.IdentifierMixin, object):
         object_filter = {"hardware": {"virtualHost": {"id": {"operation": "not null"}}}}
         mask = "mask[virtualHost[guests[powerState]]]"
         return self.client.call('SoftLayer_Account', 'getHardware', mask=mask, filter=object_filter)
+
+    def authorize_storage(self, vs_id, username_storage):
+        """Authorize File or Block Storage to a Virtual Server.
+
+        :param int vs_id: Virtual server id.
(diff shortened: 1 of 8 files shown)


SoftLayer/managers/vs.py
```diff
@@ -1404,3 +1404,45 @@ def get_hardware_guests(self):
...
+        """Attach portal storage to a Virtual Server.
+
+        :param int vs_id: Virtual server id.
+        :param int portable_id: Portal storage id.
+
+        :return: SoftLayer_Provisioning_Version1_Transaction.
+        """
+        disk_id = portable_id
+        result = self.client.call('SoftLayer_Virtual_Guest', 'attachDiskImage',
+                                  disk_id, id=vs_id)
```
caberos, line 1446: result = self.client.call('SoftLayer_Virtual_Guest', 'attachDiskImage', portable_id, id=vs_id)
is not better this?

allmightyspiff (approved):
Looks good, thanks

SoftLayer/managers/vs.py
allmightyspiff: I likely would have written `self.client.call('SoftLayer_Virtual_Guest', 'attachDiskImage', portable_id, id=vs_id)` as I think it looks better, taking up less lines.

But there are plenty of cases in the code that would break the call up like this due to line length, so this pattern of breaking up the call into 2 lines after the method is fine too.

Its not something I feel strongly enough about to make you change though.

Issue #1445, opened 2021-03-24
add 'Billing' and 'lastTransaction' on slcli hardware detail


caberos:
add datatypes on slcli command simulate as an IBMCloud page

PR #1446, opened 2021-03-24
add billing and lastTransaction on hardware detail

caberos:
https://github.com/softlayer/softlayer-python/issues/1445

### DIFF
--- a/SoftLayer/CLI/hardware/detail.py
+++ b/SoftLayer/CLI/hardware/detail.py
@@ -61,8 +61,11 @@ def cli(env, identifier, passwords, price):
     table.add_row(['os_version', operating_system.get('version') or formatting.blank()])
     table.add_row(['created', result['provisionDate'] or formatting.blank()])
     table.add_row(['owner', owner or formatting.blank()])
-    table.add_row(['last_transaction',
-                   utils.lookup(result, 'lastTransaction', 'transactionGroup', 'name')])
+
+    last_transaction = "{} ({})".format(utils.lookup(result, 'lastTransaction', 'transactionGroup', 'name'),
+                                        utils.clean_time(utils.lookup(result, 'lastTransaction', 'modifyDate')))
+
+    table.add_row(['last_transaction', last_transaction])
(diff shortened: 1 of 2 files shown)


SoftLayer/CLI/hardware/detail.py
```diff
@@ -61,6 +61,8 @@ def cli(env, identifier, passwords, price):
     table.add_row(['os_version', operating_system.get('version') or formatting.blank()])
     table.add_row(['created', result['provisionDate'] or formatting.blank()])
     table.add_row(['owner', owner or formatting.blank()])
+    table.add_row(['LastTransaction', result['lastTransaction']['transactionGroup']['name']])
```
ATGE, line 64:
```suggestion
   table.add_row(['last_transaction', utils.lookup(result, 'lastTransaction', 'transactionGroup', 'name')])
```

allmightyspiff: If were going to add the last transaction, I  think it would be helpful to know when it was run. So something like this:

```
last_transaction = "{} ({}}".format(utils.lookup(result, 'lastTransaction', 'transactionGroup', 'name'), 
                                                     utils.clean_time(utils.lookup(result, 'lastTransaction', 'modifyDate'))
table.add_row(['last_transaction', utils.lookup(result, 'lastTransaction', 'transactionGroup', 'name')])
```

That way we don't have to add an extra row in the table. Should end up looking like

`Service Setup (2021-01-01 12:15)`

allmightyspiff (changes requested):
Add the date to the transaction output, and use `utils.lookup` incase you get a server that doesn't have a last transaction for some wild reason.

ATGE (approved):
### Testing

- ✔️ `$ slcli hw detail 1403539 | grep last_transaction`
`last_transaction  CENT OS Reload (2020-07-24T10:12:00-06:00)  `     

looks good

FernandoOjeda (approved):
Looks good for me.

allmightyspiff (approved):

PR #1460, opened 2021-04-08
Fixes shift+ins when pasteing into a password field for windows users.

### LINKED ISSUE #1436: slcli setup in windows puts in bad whitespace with ctrl+v
running `slcli setup` in command.exe in windows seems to be getting bad characters in when using ctrl+v to input API keys. Using a mouse right-click seems to work ok.

Need to make sure we trim out any new lines, white space, etc.

allmightyspiff:
Fixes #1436

`getpass()` is only ever called in `slcli config setup`, so that command is the only one that needs to be tested.

The problem specifically is with windows environments (powershell and cmd.exe) and shift+insert. When inputting a password, the user should be able to use shift+insert to paste their API key/password. Previously, shift+insert would put the character sequence `àR`, so this fix looks for that sequence, then reads from the clipboard. 

Shift+insert works just fine with `input()`, its only when the text is hidden does this problem occur.

### DIFF
--- a/SoftLayer/CLI/environment.py
+++ b/SoftLayer/CLI/environment.py
@@ -67,7 +67,22 @@ class Environment(object):
 
     def getpass(self, prompt, default=None):
         """Provide a password prompt."""
-        return click.prompt(prompt, hide_input=True, default=default)
+        password = click.prompt(prompt, hide_input=True, default=default)
+
+        # https://github.com/softlayer/softlayer-python/issues/1436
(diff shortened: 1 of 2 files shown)


ATGE (approved):
### Testing 
- ✔️ slcli config setup
 cmd.exe 
looks good!

caberos (approved):
LGTM.

FernandoOjeda (approved):
Looks good for me.

Issue #1472, opened 2021-04-21
add the firewall information on slcli firewall detail


caberos:
add the firewall information data as the portal show the data

PR #1475, opened 2021-04-22
add the firewall information on slcli firewall detail

caberos:
https://github.com/softlayer/softlayer-python/issues/1472


SoftLayer/managers/firewall.py
```diff
@@ -290,3 +290,15 @@ def edit_standard_fwl_rules(self, firewall_id, rules):
...
 
         return rule_svc.createObject(template)
+
+    def get_instance(self, firewall_id, mask=None):
+        """Get the firewall information
+
+        :param integer firewall_id: the instance ID of the standard firewall
+        """
+        if not mask:
+            mask = ('mask[datacenter,networkVlan]')
```
allmightyspiff, line 300: This should just be a string, remove the `()`

tests/CLI/modules/firewall_tests.py
```diff
@@ -59,27 +59,32 @@ def test_detail(self):
         result = self.run_command(['firewall', 'detail', 'vlan:1234'])
         self.assert_no_fail(result)
         self.assertEqual(json.loads(result.output),
-                         [{'#': 1,
```
allmightyspiff, line 62: You dont need to test the entire output be exact.

My suggestion would be to do something like this:

```
json_result = json.loads(result.output)
self.assertEqual(json_result[0]['action'], 'permit')
```

for whatever fields you want to test for, usually just doing a few spot checks is fine.

SoftLayer/CLI/firewall/detail.py
```diff
@@ -19,12 +19,24 @@ def cli(env, identifier):
     mgr = SoftLayer.FirewallManager(env.client)
 
     firewall_type, firewall_id = firewall.parse_id(identifier)
+    result = mgr.get_instance(firewall_id)
+
+    table = formatting.KeyValueTable(['name', 'value'])
+    table.align['name'] = 'r'
+    table.align['value'] = 'l'
+
+    table.add_row(['id', result['id']])
```
allmightyspiff, line 28: use either `result.get()` or `utils.lookup()` (for the nested dictionary items) for all of these add_row entries. If, for some strange reason, these fields are not returned from the API they cause hard crashes of the slcli which we want to avoid.

allmightyspiff (changes requested):
A few minor changes needed, and please merge in master to this branch to get the build to pass nicely.

Thanks.

ATGE (commented):
### Testing
-✔️
 ```bash
slcli firewall detail vlan:3080
:..................:.............................................................................................:
:             name : value                                                                                       :
:..................:.............................................................................................:
:               id : 3080                                                                                        :
: primaryIpAddress : 5.153.29.130                                                                                :
:       datacenter : Amsterdam 1                                                                                 :
:      networkVlan : FirewallTesting                                                                             :
:    networkVlaniD : 361652                                                                                      :
:            rules : :...:........:..........:...............:.................:.............:.................: :
:                  : : # : action : protocol :     src_ip    :     src_mask    :     dest    :    dest_mask    : :
:                  : :...:........:..........:...............:.................:.............:.................: :
:                  : : 1 : permit :   tcp    :    0.0.0.0    :     0.0.0.0     : any:1-65535 : 255.255.255.255 : :
:                  : : 2 : permit :   udp    :    0.0.0.0    :     0.0.0.0     : any:1-65535 : 255.255.255.255 : :
:                  : : 3 : permit :   tcp    : 50.23.236.228 : 255.255.255.255 : any:100-200 : 255.255.255.255 : :
:                  : : 4 : permit :   tcp    :    0.0.0.0    :     0.0.0.0     :  any:80-80  : 255.255.255.255 : :
:                  : :...:........:..........:...............:.................:.............:.................: :
:..................:.............................................................................................:

```
Looks good, just keep in mind add a rules type(current rules just displayed are ipv4, UI has ipv6 too), maybe add router (as UI shows ) field too.

ATGE (approved):

FernandoOjeda (approved):
Looks good.

allmightyspiff (approved):

PR #1477, opened 2021-04-26
Add the reserved capacity option to the slcli.

FernandoOjeda:
Add the reserved capacity option to the slcli.


SoftLayer/CLI/reservedcapacity/vs_instances.py
```diff
@@ -0,0 +1,39 @@
...
+
+    guest_table = formatting.Table([
+        'status', 'name', 'location', 'cpu', 'ram'
+    ])
+
+    for instance in reserved_capacity_result:
+        if 'guest' in instance:
+            ram = utils.lookup(instance, 'guest', 'maxMemory')
+            ram_bytes = ram * (1024 ** 2)
+            ram_result = ram_bytes / (1024 ** 3)
```
caberos, line 30: `1024` I think it should be in a class of fixed variables and import them from there

allmightyspiff (changes requested):
So, reserved capacity is already a part of the SLCLI.

For the CLI, these commands are under `SoftLayer/CLI/virt/capacity/` and for the manager its under `SoftLayer/managers/vs_capacity.py`

Your edit and vs_instances commands will still be useful, just need to be moved to the correct area, and use the already existing manager class.

Issue #1489, opened 2021-05-18
slcli vlan cancel should report if a vlan is automatic


allmightyspiff:
When cancelling a vlan with `slcli vlan cancel` we should change the messaging a bit  to be a bit more clear to users why certain vlans cannot be cancelled.

If the VLAN doesn't have a billing Item, meaning it is an automatically assigned vlan, and cannot be cancelled. Error should be
> VLAN <number> is an automatically assigned and free of charge VLAN, it will automatically be removed from your account when it is empty

PR #1495, opened 2021-05-27
slcli vlan cancel should report if a vlan is automatic

caberos:
https://github.com/softlayer/softlayer-python/issues/1489


SoftLayer/CLI/vlan/cancel.py
```diff
@@ -0,0 +1,30 @@
...
+@click.argument('identifier')
+@environment.pass_env
+def cli(env, identifier):
+    """Cancel network vlan."""
+
+    mgr = SoftLayer.NetworkManager(env.client)
+    billing = BillingManager(env.client)
+    if not (env.skip_confirmations or formatting.no_going_back(identifier)):
+        raise exceptions.CLIAbort('Aborted')
+
```
ATGE, line 23: adding a resolve vlan name as `helpers.resolve_id(manager.resolve_ids, volume_id, 'Network Vlan')` would be nice

SoftLayer/managers/billing.py
```diff
@@ -0,0 +1,27 @@
+"""
```
ATGE, line 1: As the mostly managed Manager modules their own billing stuff, maybe it's better to move this logic into the Network Manager, but in any case, keep in mind to add tests for that.

FernandoOjeda (commented):
Do not forget to add unit test for the new methods implemented.

SoftLayer/CLI/vlan/cancel.py
allmightyspiff: The problem with a resolver here is that vlan Number and vlan Name are both not unique fields, so you might accidently cancel a vlan you didn't mean to. Lets just stick with vlan ID here.

SoftLayer/CLI/vlan/cancel.py
```diff
@@ -0,0 +1,30 @@
...
+    item = mgr.get_vlan(identifier).get('billingItem')
+    if item:
+        billing.cancel_item(item.get('id'), 'cancel by cli command')
```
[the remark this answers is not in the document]
allmightyspiff: Yes, we can let the customer add a reason, but having the default be "Cancelled from SLCLI" is good to. I know that I would hate to have to put a cancellation reason everytime, and I think most customers don't care what the reason is.

SoftLayer/managers/network.py
```diff
@@ -752,3 +753,10 @@ def set_subnet_ipddress_note(self, identifier, note):
         """
         result = self.client.call('SoftLayer_Network_Subnet_IpAddress', 'editObject', note, id=identifier)
         return result
+
+    def get_cancel_failure_reasons(self, identifier):
+        """get the reasons by cannot cancel the VLAN
```
allmightyspiff, line 758:
```suggestion
        """get the reasons why we cannot cancel the VLAN
```

allmightyspiff (changes requested):
Change the logic of CLI/vlan/cancel.py to something like this:



```
reasons = mgr.get_cancel_failure_reasons(identifier)
if len(reasons) > 0:
    raise exceptions.CliAbort(reasons )
item = mgr.get_vlan(identifier).get('billingItem')
if item:
    billing.cancel_item(item.get('id'), 'cancel by cli command')
else:
    raise exceptions.CliAbort("VLAN is an automatically assigned and free of charge VLAN, it will automatically be removed from your account when it is empty")

```

allmightyspiff (approved):
Changes look good, thanks.

FernandoOjeda (approved):
Looks good.

SoftLayer/CLI/vlan/cancel.py
```diff
@@ -0,0 +1,35 @@
...
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import exceptions
+from SoftLayer.CLI import formatting
+
+
+@click.command()
+@click.argument('identifier')
+@environment.pass_env
+def cli(env, identifier):
+    """Cancel network vlan."""
```
ATGE, line 16:
```suggestion
    """Cancel a Network VLAN."""
```

ATGE (approved):
### Testing
- ✔️slcli  vlan cancel 1234556
- ✔️slcli -v -y vlan cancel 1234556
- ✔️slcli vlan cancel -h' 
-✔️ slcli vlan -h

looks good, just a comment on cancel cli method docs.

Issue #1505, opened 2021-06-15
fix to SLCLI table look well, new method to fix the table disconfigurate when the text data is very long


caberos:
new method to fix the table disconfigurate 
![image](https://user-images.githubusercontent.com/22535437/122117780-41d39900-cdf5-11eb-9db7-32f5c20ec476.png)

PR #1506, opened 2021-06-15
new method to fix the table disconfigurate when the text data is very…

caberos:
https://github.com/softlayer/softlayer-python/issues/1505

### DIFF
--- a/SoftLayer/utils.py
+++ b/SoftLayer/utils.py
@@ -408,11 +408,11 @@ def trim_to(string, length=80, tail="..."):
         return string
 
 
-def format_comment(comment, max_line_length):
+def format_comment(comment, max_line_length=60):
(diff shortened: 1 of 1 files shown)


SoftLayer/utils.py
```diff
@@ -406,3 +406,24 @@ def trim_to(string, length=80, tail="..."):
         return string[:length] + tail
     else:
         return string
+
+
+def format_comment(comment, max_line_length):
```
allmightyspiff, line 411: Set a default value for `max_line_length`, maybe 60 characters here would be good as a default.

allmightyspiff (changes requested):
Looks good, just set a default for max_line_length.

allmightyspiff (approved):

PR #1514, opened 2021-07-06
Updates to the Release workflow for publishing to test pypi

allmightyspiff:
This updates the `release` Job to also publish the package to test-pypi repo, and if that works for the next few versions, I'll update it to do the official one as well.

I was going to try to get a way to auto-publish a package every update to master, but I think that will end up being too much trouble for now.

Any dev releases we do can be easily done by just adding a -devX suffix to the version string I think and push that to the test-pypi branch.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,4 @@
-name: Release
+name: Release Snapcraft and PyPi (Testing)
 
 on:
   release:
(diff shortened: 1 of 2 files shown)


ATGE (approved):
looks good

FernandoOjeda (approved):
Looks good.

PR #1517, opened 2021-07-07
Adding in CodeQL Analysis

allmightyspiff:
This should add a new job that will scan the code for security issues, so well see how this works.


ATGE (approved):
looks good

Issue #1520, opened 2021-07-19
Refactor `loadbal order-options`


allmightyspiff:
`slcli lb order-options` takes a really long time, and produces not great output.

`slcli lb order-options` without arguments should just list out datacenters, and a message at the bottom letting the user know they should select a datacenter to view pricing and subnets

`slcli lb order-options -d dal10` works fine, but should output 2 tables, instead of 1 table with a table in each column.

Table 1 titled "<Datacenter> Prices", Table 2 titled "<Datacenter> Subnets Private Subnets"

allmightyspiff:
Current output:

```
$ slcli --format=table lb order-options -d dal10
Prints options for ordering
:..............................................................................................:
:                                  DALLAS10: DAL10 - Dallas                                    :
:.........................................:....................................................:
:                  Prices                 :                  Private Subnets                   :
:.........................................:....................................................:
: :..............................:......: : :.........:...................:..................: :
: :           KeyName            : Cost : : :    Id   :       Subnet      :       Vlan       : :
: :..............................:......: : :.........:...................:..................: :
: :  LOAD_BALANCER_AS_A_SERVICE  :  0   : : : 1596063 :  10.93.138.192/26 : dal10.pod03.886  : :
: :     LOAD_BALANCER_UPTIME     : .025 : : : 1301433 : 10.177.122.168/29 : dal10.pod01.1251 : :
: :   LOAD_BALANCER_BANDWIDTH    :  0   : : : 1466083 :  10.177.211.0/26  : dal10.pod01.1251 : :
: : LOAD_BALANCER_DATA_PROCESSED :  0   : : : 2340712 :  10.221.68.192/26 : dal10.pod01.777  : :
: :..............................:......: : :.........:...................:..................: :
:.........................................:....................................................:
```

Proposed output: (something like this anyway)
![image](https://user-images.githubusercontent.com/7408017/126829547-5a523a37-8bcd-4be6-805d-d854c7d85569.png)

PR #1521, opened 2021-07-26
Refactor loadbal order-options

caberos:
https://github.com/softlayer/softlayer-python/issues/1520

### DIFF
--- a/SoftLayer/CLI/loadbal/order.py
+++ b/SoftLayer/CLI/loadbal/order.py
@@ -89,53 +89,64 @@ def order_options(env, datacenter):
     net_mgr = SoftLayer.NetworkManager(env.client)
     package = mgr.lbaas_order_options()
 
-    for region in package['regions']:
-        dc_name = utils.lookup(region, 'location', 'location', 'name')
+    if not datacenter:
+        data_table = formatting.KeyValueTable(['Datacenters', 'City'])
+        for region in package['regions']:
+            data_table.add_row([region['description'].split('-')[0], region['description'].split('-')[1]])
+            # print(region)
+        env.fout(data_table)
+        click.secho("ERROR: Use `slcli lb order-options --datacenter <DC>` "
+                    "to find pricing information and private subnets for that specific site.")
 
-        # Skip locations if they are not the one requested.
-        if datacenter and dc_name != datacenter:
-            continue
-        this_table = formatting.Table(
-            ['Prices', 'Private Subnets'],
-            title="{}: {}".format(region['keyname'], region['description'])
-        )
+    else:
+        for region in package['regions']:
+            dc_name = utils.lookup(region, 'location', 'location', 'name')
 
-        l_groups = []
-        for group in region['location']['location']['groups']:
-            l_groups.append(group.get('id'))
-
-        # Price lookups
-        prices = []
-        price_table = formatting.KeyValueTable(['KeyName', 'Cost'])
-        for item in package['items']:
-            i_price = {'keyName': item['keyName']}
-            for price in item.get('prices', []):
-                if not price.get('locationGroupId'):
-                    i_price['default_price'] = price.get('hourlyRecurringFee')
-                elif price.get('locationGroupId') in l_groups:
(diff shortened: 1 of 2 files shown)


ATGE (commented):
the command `slcli lb order-options` works well, just the message displayed as an error looks weird.
Also, the output of the `slcli lb order-options --datacenter <DC>` in the worked branch is the same output as the latest `slcli` version shows, so looks like the table requirement mentioned in the issue, is not implemented. #1520 

> `slcli lb order-options -d dal10` works fine, but should output 2 tables, instead of 1 table with a table in each column.
> 
> Table 1 titled " Prices", Table 2 titled " Subnets Private Subnets"
### Testing
 - ✔️ slcli loadbal order-options
 
 
image

- slcli loadbal order-options -d dal13
 worked branch output:
image
 slcli (SoftLayer Command-line) v5.9.6 output:
image

SoftLayer/CLI/loadbal/order.py
```diff
@@ -89,53 +89,64 @@ def order_options(env, datacenter):
...
+    if not datacenter:
+        data_table = formatting.KeyValueTable(['Datacenters', 'City'])
+        for region in package['regions']:
+            data_table.add_row([region['description'].split('-')[0], region['description'].split('-')[1]])
+            # print(region)
+        env.fout(data_table)
+        click.secho("ERROR: Use `slcli lb order-options --datacenter <DC>` "
+                    "to find pricing information and private subnets for that specific site.")
+
+    else:
```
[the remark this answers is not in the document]
allmightyspiff: Most of this logic is just table formatting, so I'm ok with it not being in a manager specifically, unless we want to use it in more than just this one command. Breaking it up into a few smaller sub-functions that return `Table` objects would make this logic a bit more readable though.

SoftLayer/CLI/loadbal/order.py
```diff
@@ -89,53 +89,64 @@ def order_options(env, datacenter):
...
+                # Only show these types, easier to filter here than in an API call.
+                if subnet.get('subnetType') != 'PRIMARY' and \
+                        subnet.get('subnetType') != 'ADDITIONAL_PRIMARY':
+                    continue
+                space = "{}/{}".format(subnet.get('networkIdentifier'), subnet.get('cidr'))
+                vlan = "{}.{}".format(subnet['podName'], subnet['networkVlan']['vlanNumber'])
+                subnet_table.add_row([subnet.get('id'), space, vlan])
+            this_table.add_row([price_table, subnet_table])
+
+            env.fout(this_table)
```
allmightyspiff, line 149: I'd like this to be 2 tables now, so the following should produce the output I'm looking for.

```suggestion
            env.fout(price_table)
            env.fout(subnet_table)
```

allmightyspiff (changes requested):
Along with my comments, I'd like to see a unit test specifically for this command with a location specified.

SoftLayer/CLI/loadbal/order.py
```diff
@@ -89,53 +89,60 @@ def order_options(env, datacenter):
...
+        env.fout(data_table)
+        click.secho("Use `slcli lb order-options --datacenter <DC>` "
+                    "to find pricing information and private subnets for that specific site.")
+
+    else:
+        for region in package['regions']:
+            dc_name = utils.lookup(region, 'location', 'location', 'name')
+
+            # Skip locations if they are not the one requested.
+            if datacenter and dc_name != datacenter:
```
allmightyspiff, line 106: The datacenter lists the DC name as all uppercase (`DAL13`), but if you search for `DAL13`, this command fails to find the right DC (as opposed to `dal13`)

Need to lowercase the input so it matches what we get from the API.

```suggestion
            if datacenter and dc_name != datacenter.lower():
```

image

allmightyspiff (changes requested):
Make sure to lowercase user input for datacenter here. Otherwise this looks great and ready to merge. Thanks.

allmightyspiff (approved):
Thanks, changes look good.

Issue #1525, opened 2021-08-02
`slcli server create-options dal13 ` Error


allmightyspiff:
```
$ slcli -v server create-options --prices dal13
Calling: SoftLayer_Product_Package::getAllObjects(id=None, mask='mask[id]', filter='{'keyName': {'operation': 'BARE_METAL_SERVER'}}', args=(), limit=None, offset=None))
Calling: SoftLayer_Product_Package::getItems(id=200, mask='mask[id,keyName,capacity,description,attributes[id,attributeTypeKeyName],itemCategory[id,categoryCode],softwareDescription[id,referenceCode,longDescription],prices[categories]]', filter='None', args=(), limit=None, offset=None))
Calling: SoftLayer_Product_Package::getActivePresets(id=200, mask='mask[prices]', filter='None', args=(), limit=None, offset=None))
Calling: SoftLayer_Product_Package::getAccountRestrictedActivePresets(id=200, mask='mask[prices]', filter='None', args=(), limit=None, offset=None))
Calling: SoftLayer_Product_Package::getRegions(id=200, mask='location[location[priceGroups]]', filter='None', args=(), limit=None, offset=None))
Calling: SoftLayer_Location::getDatacenters(id=None, mask='mask[priceGroups]', filter='{'name': {'operation': 'dal13'}}', args=(), limit=1, offset=None))
Calling: SoftLayer_Account::getRouters(id=None, mask='dal13', filter='', args=(), limit=None, offset=None))
SoftLayerAPIError(500): Property 'dal13' not valid for 'SoftLayer_Hardware'.
```

https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/CLI/hardware/create_options.py#L24 Needs to be 

```
 routers = account_manager.get_routers(None, location)
```

Currently, the location is getting passed in as a mask, which is wrong.

allmightyspiff:
Make sure to add a unit test for this specific case as well.

PR #1526, opened 2021-08-02
slcli server create-options dal13 Error

caberos:
https://github.com/softlayer/softlayer-python/issues/1525


SoftLayer/managers/account.py
```diff
@@ -277,7 +277,7 @@ def get_account_all_billing_orders(self, limit=100, mask=None):
         return self.client.call('Billing_Order', 'getAllObjects',
                                 limit=limit, mask=mask)
 
-    def get_routers(self, mask=None, location=None):
+    def get_routers(self, location=None, mask=None):
```
FernandoOjeda, line 280: I suggest you to fix it direct in the cli hardware create-options module, like this:
```python
routers = account_manager.get_routers(location=location)
```

allmightyspiff: agreed, being specific about parameters can help avoid bugs like this in the future.

allmightyspiff (changes requested):

allmightyspiff (changes requested):
I think you accidently reverted your change to `SoftLayer/managers/account.py`, can you put the function back to `def get_routers(self, location=None, mask=None):` and this will be good to go. Sorry if there was any confusion about that.

allmightyspiff (approved):
Looks good. thanks.

PR #1527, opened 2021-08-02
Prevents SLCLI_VERSION environment variable from breaking things

### LINKED ISSUE #1524: Error: Invalid value for '--version': '5.9.6' is not a valid boolean.
> Reminder: No username or APIkeys should be added to these issues, as they are public.

**Describe the bug**
A clear and concise description of what the bug is.  Include the command you used, make sure to include the `-v` flag, as that information is very helpful. Ex: `slcli -v vs list`

I'm building a Dockerfile using registry.access.redhat.com/ubi8/ubi and python 3.9 (i've also tried 3.8).  I've installed softlayer multiple ways, all with the same results:
1.  python -m pip install softlayer
2.  python setup.py install

If I run the `slcli` with no options it returns the help menu, no errors.  If I run the `slcli -v` or some of the other options, it fails with `Error: Invalid value for '--version': '5.9.6' is not a valid boolean.` at the bottom of the command.

I've tried installing older versions of softlayer but the same error, with the same version 5.9.6 would still show.  I'm guessing because that version is considered the latest in the --version command.

**Expected behavior**

The error should not appear.

**Screenshots**

```
slcli -v
Usage: slcli [OPTIONS] COMMAND [ARGS]...
Try 'slcli -h' for help.

Error: Invalid value for '--version': '5.9.6' is not a valid boolean.
```

**Version**
Include the output of `slcli --version`

```
slcli --version
Current: slcli (SoftLayer Command-line) v5.9.6
Latest:  slcli (SoftLayer Command-line) v5.9.6
```

allmightyspiff:
Fixes #1524 

Apparently Click has a neat feature that allows you to set parameters with environment variables, which I just learned about.
https://github.com/softlayer/softlayer-python/blame/master/SoftLayer/CLI/core.py#L103 is what actually controls that, but since its been active for 3 years, I don't really want to turn it off entirely, as it is a somewhat neat feature.

So I'll just turn it off for --version which I don't think anyone would want to have automatically be set.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,7 +5,7 @@ on:
     types: [published]
 
 jobs:
-  release:
+  snap-release:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
--- a/README.rst
+++ b/README.rst
@@ -76,6 +76,26 @@ InsecurePlatformWarning Notice
 ------------------------------
 This library relies on the `requests <http://docs.python-requests.org/>`_ library to make HTTP requests. On Python versions below Python 2.7.9, requests has started emitting a security warning (InsecurePlatformWarning) due to insecurities with creating SSL connections. To resolve this, upgrade to Python 2.7.9+ or follow the instructions here: http://stackoverflow.com/a/29099439.
 
+Basic Usage
+-----------
+
+- `The Complete Command Directory <https://softlayer-python.readthedocs.io/en/latest/cli_directory/>`_
+
+Advanced Usage
+--------------
+
+You can automatically set some parameters via environment variables with by using the SLCLI prefix. For example
+
+.. code-block:: bash
+  $ export SLCLI_VERBOSE=3
+  $ export SLCLI_FORMAT=json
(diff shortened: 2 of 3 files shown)


allmightyspiff:
Also updated the snapcraft release build to have a unique name since I want to use `release` for something else in the future.
Updated the README to highlight the environment variable feature.

caberos (commented):
looks good

**Testing**
 slcli --version
Current: slcli (SoftLayer Command-line) v5.9.5
Latest:  slcli (SoftLayer Command-line) v5.9.6

slcli -v
Usage: slcli [OPTIONS] COMMAND [ARGS]...
Try "slcli -h" for help.

Error: Missing command.

slcli -v vs list
Calling: SoftLayer_Account::getVirtualGuests(id=None, mask='primaryIpAddress, activeTransaction[ id,transactionStatus[name,friendlyName] ],hostname,datacenter.name,id,primaryBackendIpAddress', filter='{}', args=(), limit=100, offset=0))

caberos (approved):

PR #1546, opened 2021-09-16
Fix account events

### LINKED ISSUE #1545: `slcli account events` shows invalid JSON
There is an erroneous print statement in  `unplanned_event_table()` https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/CLI/account/events.py#L72 that needs to be removed

ATGE:
Fixes #1545 

### Testing
- ✔️ slcli --format json account events


allmightyspiff (approved):
Looks good, dont worry about the `analysis` errors, those are new and I will likely  set them to be ignored I think.

PR #1554, opened 2021-10-21
Snapshot notify

hariha74:
Function change to enable or disable notification for snapshot notification


allmightyspiff (approved):
Need to fix the code style errors first, but otherwise looks fine. You can check the analysis with `tox -e analysis` in your local softlayer-python directory.

Issue #1577, opened 2022-01-24
Bandwidth pool management


allmightyspiff:
I'm currently adding `slcli account bandwidth-pools` in #1575 but we could use a way to add/remove servers from the pool along with a more detailed view of what is in the pool. Basically copy https://cloud.ibm.com/classic/network/bandwidth/vdr/309961

PR #1582, opened 2022-02-08
Bandwidth pool management

caberos:
https://github.com/softlayer/softlayer-python/issues/1577

### DIFF
--- a/SoftLayer/CLI/account/bandwidth_pools_detail.py
+++ b/SoftLayer/CLI/account/bandwidth_pools_detail.py
@@ -1,0 +1,66 @@
+"""Get bandwidth pools."""
+# :license: MIT, see LICENSE for more details.
+import click
+
+from SoftLayer import AccountManager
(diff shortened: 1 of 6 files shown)


BrianSantivanez:
Running the help command
image

Running the command with an id that does not exist
image

Id´s used to test
image

Running the command with first id
image

Running the command with a second id
image

edsonarios:
Test new feature
image
image
image

Maybe it's necessary to see the id of the results, because when I want to see the detail of a region, I need the id
image

allmightyspiff:
> For found the id of the any region, maybe we need the id of the follow command "slcli account bandwidth-pools" but in this commando not show us the id
I created #1585  to address this.

BrianSantivanez:
image

When command is run with 1092822 and 138442 is falling

image

image

edsonarios (approved):
Looks good to me

allmightyspiff (approved):

PR #1584, opened 2022-02-09
Replace the use of ptable with prettytable

### LINKED ISSUE #1583: Ptable is abandoned and can be replaced with prettytable
**Describe the bug**
The [ptable](https://github.com/kxxoling/PTable) dependency is abandoned and can be replaced with [prettytable](https://github.com/jazzband/prettytable/), which is actively maintained.

Recently prettytable started to occupy the same namespace as ptable, which means that they can not be installed side-by-side in a packaging context anymore (and they shouldn't be either).

**Expected behavior**
Have a maintained version of prettytable.

**Screenshots**
n/a

**Version**
all

dvzrv:
{README.rst,tools/*}:
Replace ptable with prettytable >= 2.0.0.

SoftLayer/CLI/formatting.py:
Only import prettytable.

Fixes #1583

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -167,7 +167,7 @@ If you cannot install python 3.6+ for some reason, you will need to use a versio
 
 Python Packages
 ---------------
-* ptable >= 0.9.2
+* prettytable >= 2.0.0
 * click >= 7
 * requests >= 2.20.0
 * prompt_toolkit >= 2
--- a/SoftLayer/CLI/formatting.py
+++ b/SoftLayer/CLI/formatting.py
@@ -11,11 +11,7 @@ import os
 
 import click
 
-# If both PTable and prettytable are installed, its impossible to use the new version
-try:
-    from prettytable import prettytable
-except ImportError:
-    import prettytable
+import prettytable
 
 from SoftLayer.CLI import exceptions
 from SoftLayer import utils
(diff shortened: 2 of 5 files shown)


allmightyspiff (changes requested):
I think you need to update `setup.py` to add this library as well. That should at least get the tests to work. 

My only concern is that anyone with the `PTable` library installed will run into import errors, which is the problem I ran into last time I tried to switch over to 100% PrettyTable (hence try/except in formatting.py).

Thanks for the pull request though, provided this doesn't cause users upgrade issues this change should be good.

dvzrv:
> I think you need to update `setup.py` to add this library as well. That should at least get the tests to work.

Ah, right! There are a bit too many places where the dependencies are set :laughing: 
Thanks for the hint!

> My only concern is that anyone with the `PTable` library installed will run into import errors, which is the problem I ran into last time I tried to switch over to 100% PrettyTable (hence try/except in formatting.py).

I guess ptable should be dropped from all systems. There has not been any development effort in years.

> Thanks for the pull request though, provided this doesn't cause users upgrade issues this change should be good.

Thanks! I hope it works out as intended, as it allows us on Arch Linux to drop the python-ptable package (which conflicts with python-prettytable and is only required for this project)

allmightyspiff (approved):
Looks good thanks.
Tested upgrading from 5.9.9 to this version ran without issue, so no worries there either.

PR #1592, opened 2022-02-28
Datacenter closure report

### LINKED ISSUE #1590: Report on closing datacenters
Add a new command `slcli report datacenter-closures` (or something similar, might come up with a better name later) that lists all pods that are closing. Use https://sldn.softlayer.com/reference/services/SoftLayer_Network_Pod/getAllObjects/ and check the capabilities property for `CLOSURE_ANOUNCED` to see if a pod is closing. A table should show each pod being closed, with hardware/VSI/vlan/subnet/anything else we can think of in that pod.

We might be able to use the SoftLayer_Search api to pull anything based off the pod it is in.

allmightyspiff:
Fixes #1590 basic structure for the DC closure report

### DIFF
--- a/SoftLayer/CLI/report/dc_closures.py
+++ b/SoftLayer/CLI/report/dc_closures.py
@@ -1,0 +1,125 @@
+"""Metric Utilities"""
+import datetime
+import itertools
+import sys
+
+import click
+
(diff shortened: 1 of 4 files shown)


caberos (approved):
LGTM

Issue #1591, opened 2022-02-28
New Command `slcli hardware|virtual monitoring` 


allmightyspiff:
Basically display the information seen here:

https://cloud.ibm.com/gen1/infrastructure/monitoring/3123796/bare-metal

![image](https://user-images.githubusercontent.com/7408017/156031902-b9b8dd86-c4a6-4839-af1f-3c3353890036.png)

Command should work for both hardware and virtual guests since they are pretty similar.

## TODO
- list monitoring
- Edit a monitor
- Delete a monitor
- Add a monitor
- Add a notified user 
- Remove a notified user

When working on this issue, each of these should be created into a new github issue for tracking/merging.

PR #1593, opened 2022-03-10
New Command slcli hardware|virtual monitoring

caberos:
https://github.com/softlayer/softlayer-python/issues/1591


SoftLayer/CLI/hardware/monitoring.py
```diff
@@ -0,0 +1,37 @@
...
+    table.align['value'] = 'l'
+
+    monitoring = hardware.get_hardware(identifier)
+
+    table.add_row(['domain', monitoring.get('fullyQualifiedDomainName')])
+    table.add_row(['public Ip', monitoring.get('primaryIpAddress')])
+    table.add_row(['private Ip', monitoring.get('primaryBackendIpAddress')])
+    table.add_row(['location', monitoring['datacenter']['longName']])
+
+    monitoring_table = formatting.Table(['Id', 'ipAddress', 'status', 'type', 'notify'])
```
allmightyspiff, line 30: Capitalize the first letter for column/key names, and "IP" should have both capitalized (at least when its a title/key) since its an acronym 
```suggestion
    table.add_row(['Domain', monitoring.get('fullyQualifiedDomainName')])
    table.add_row(['Public IP', monitoring.get('primaryIpAddress')])
    table.add_row(['Private IP', monitoring.get('primaryBackendIpAddress')])
    table.add_row(['Location', monitoring['datacenter']['longName']])

    monitoring_table = formatting.Table(['Id', 'IP Address', 'Status', 'Type', 'Notify'])
```

SoftLayer/CLI/virt/monitoring.py
```diff
@@ -0,0 +1,37 @@
...
+    table.align['value'] = 'l'
+
+    monitoring = vsi.get_instance(identifier)
+
+    table.add_row(['domain', monitoring.get('fullyQualifiedDomainName')])
+    table.add_row(['public Ip', monitoring.get('primaryIpAddress')])
+    table.add_row(['private Ip', monitoring.get('primaryBackendIpAddress')])
+    table.add_row(['location', monitoring['datacenter']['longName']])
+
+    monitoring_table = formatting.Table(['Id', 'ipAddress', 'status', 'type', 'notify'])
```
allmightyspiff, line 30: Same comment about capitalization as for hardware monitoring.

SoftLayer/CLI/virt/monitoring.py
```diff
@@ -0,0 +1,37 @@
+"""Get monitoring for a vSI device."""
```
allmightyspiff, line 1:
```suggestion
"""Get monitoring for a VSI device."""
```

SoftLayer/managers/hardware.py
```diff
@@ -269,6 +270,7 @@ def get_hardware(self, hardware_id, **kwargs):
                 'hourlyBillingFlag,'
                 'tagReferences[id,tag[name,id]],'
                 'networkVlans[id,vlanNumber,networkSpace],'
+                'monitoringServiceComponent,networkMonitors[queryType,lastResult,responseAction],'
```
allmightyspiff, line 273: I think I would prefer you define this mask in the command itself, and send it in as an argument. The mask in here is getting fairly lengthy already and I would like to avoid growing it further.

SoftLayer/managers/vs.py
```diff
@@ -252,6 +252,7 @@ def get_instance(self, instance_id, **kwargs):
                 'tagReferences[id,tag[name,id]],'
                 'networkVlans[id,vlanNumber,networkSpace],'
                 'dedicatedHost.id,'
+                'monitoringServiceComponent,networkMonitors[queryType,lastResult,responseAction],'
```
allmightyspiff, line 255: I think I would prefer you define this mask in the command itself, and send it in as an argument. The mask in here is getting fairly lengthy already and I would like to avoid growing it further.

allmightyspiff (changes requested):
Code looks good overall, just some slight formatting issues. And have each command define and send in its own object mask instead of adding it to the default mask please.

Thanks

BrianSantivanez:
Running the help command
image

Running the command with a hardware
image

Running the command with a virtual
image

Testing the command with a hardware and virtual guest with 3123796 and 127061328 as ids respectively is falling

allmightyspiff (approved):
code looks good, thanks

BrianSantivanez (approved):
> Try running `python setup.py install` to see if that fixes it. If not, check to make sure `ptable` is NOT installed. `ptable` was removed in the most recent version as a dependency since its a very outdated package.

Thanks @allmightyspiff , the problem was that I had `prettytable` and `ptable` installed at the same time.

After correct the observations:
Running the command with a hardware
image

Running the command with a virtual
image

Looks good to me

PR #1594, opened 2022-03-10
fix to errors in slcli hw create-options

caberos:
https://github.com/softlayer/softlayer-python/issues/1588


allmightyspiff (approved):
lgtm

Issue #1598, opened 2022-03-22
Add an `orderBy` filter to `slcli vlan list`


allmightyspiff:
SoftLayer_Account::getNetworkVlans can have the vlans in different orders between queries (see screenshot), so we need to make sure we have an orderBy filter (by the ID field) to ensure each API call has the same order between resultLimits.

![image](https://user-images.githubusercontent.com/7408017/159530014-03c830b5-e4d4-47a3-b224-2573a2cdabe1.png)

PR #1599, opened 2022-03-23
Add an orderBy filter to slcli vlan list

caberos:
https://github.com/softlayer/softlayer-python/issues/1598

### DIFF
--- a/SoftLayer/managers/network.py
+++ b/SoftLayer/managers/network.py
@@ -542,7 +542,10 @@ class NetworkManager(object):
             kwargs['mask'] = DEFAULT_VLAN_MASK
 
         kwargs['iter'] = True
-        return self.account.getNetworkVlans(mask=kwargs['mask'], filter=_filter.to_dict(), limit=limit)
+        if limit > 0:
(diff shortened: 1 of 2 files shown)


edsonarios:
Screenshots of the result 
image

edsonarios (approved):
Looks good to me

allmightyspiff (changes requested):
The only problem I have with this change is that it changes the default behavior of `list_vlans` from  "Get all vlans in pages of 100" to "Get only 100 vlans", which I think will cause issues with some customers who have a lot of vlans.

I think we should default `limit` to `None` in the function definition and in the CLI argument list. So that if the user doesn't set a limit, they get all vlans back.
If a user does set a limit, they only get that number back.

Something like this:
```python
 if limit > 0:
    return self.account.getNetworkVlans(mask=kwargs['mask'], filter=_filter.to_dict(), limit=limit)
else:
    self.account.getNetworkVlans(mask=kwargs['mask'], filter=_filter.to_dict(), iter=True)

```

allmightyspiff (approved):

Issue #1587, opened 2022-02-23
Add a warning if user orders in a POD that is being closed


allmightyspiff:
When ordering, with `slcli order place`, `slcli vs create` or `slcli hw create` emit a warning message letting the user know the pod they chose is going to be closed at some point.

You can tell if a pod is going to be closed if `CLOSURE_ANNOUNCED` property is in the capabilities list

```
$  slcli -vvv  call-api SoftLayer_Network_Pod getObject --id="mex01.pod01" --mask="mask[name,capabilities]"
Calling: SoftLayer_Network_Pod::getObject(id=mex01.pod01, mask='mask[name,capabilities]', filter='{}', args=(), limit=None, offset=None))
Starting new HTTPS connection (1): api.softlayer.com:443
https://api.softlayer.com:443 "GET /rest/v3.1/SoftLayer_Network_Pod/mex01.pod01/getObject.json?objectMask=mask%5Bname%2Ccapabilities%5D HTTP/1.1" 200 157
Returned Data:
{'capabilities': ['SUPPORTS_MULTI_VLAN_DEDICATED_FIREWALL', 'SUPPORTS_SECURITY_GROUP', 'CLOSURE_ANNOUNCED', 'DATACENTER_CLOSURE_ANNOUNCED'], 'name': 'mex01.pod01'}
:..............:............................................:
:         name : value                                      :
:..............:............................................:
: capabilities : :........................................: :
:              : :                 value                  : :
:              : :........................................: :
:              : : SUPPORTS_MULTI_VLAN_DEDICATED_FIREWALL : :
:              : :        SUPPORTS_SECURITY_GROUP         : :
:              : :           CLOSURE_ANNOUNCED            : :
:              : :      DATACENTER_CLOSURE_ANNOUNCED      : :
:              : :........................................: :
:         name : mex01.pod01                                :
:..............:............................................:

```

Given how complex ordering is, I think we should use the receipt from the order to figure out what POD the server was placed in before displaying a warning I think.

Workflow should be like this.

1. Order server normally
2. Check reciept, figure out which Network_Pod the server was placed in
3. Call `SoftLayer_Network_Pod::getObject(id=mex01.pod01, mask='mask[name,capabilities]')` to figure out if the pod is going to be closed
4. If the pod is going to be closed, send a warning (in orange/yellow) to the user about it. Otherwise do nothing.

PR #1600, opened 2022-03-28
Add a warning if user orders in a POD that is being closed

caberos:
https://github.com/softlayer/softlayer-python/issues/1587

### DIFF
--- a/SoftLayer/CLI/order/place.py
+++ b/SoftLayer/CLI/order/place.py
@@ -8,6 +8,7 @@ import click
 from SoftLayer.CLI import environment
 from SoftLayer.CLI import exceptions
 from SoftLayer.CLI import formatting
+from SoftLayer.managers import NetworkManager
 from SoftLayer.managers import ordering
 
 COLUMNS = ['keyName',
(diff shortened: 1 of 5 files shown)


BrianSantivanez:
Running with virtual guests
image

Running with hardware
image

please add the warning to `place order`

After add  to `place order` and it passes all checks it is ok to me

SoftLayer/CLI/order/place.py
```diff
@@ -90,6 +95,10 @@ def cli(env, package_keyname, location, preset, verify, billing, complex_type,
             ])
 
     else:
+        print(args)
```
allmightyspiff, line 98: Looks like this should be removed

allmightyspiff (approved):
Code looks good, just need to fix the unit tests. Thanks
And remove that one errant print(args) line

allmightyspiff (approved):

allmightyspiff:
@jonlowe-ibm I've got an issue open about it  https://github.com/softlayer/softlayer-python/issues/1687
Will be fixed in the next version. For now they can just ignore the warning if it doesn't have a datacenter listed in it.

Thanks for letting us know

Issue #1604, opened 2022-03-30
`slcli licenses` is missing the help text


allmightyspiff:
![image](https://user-images.githubusercontent.com/7408017/160909054-464649c3-70d9-4450-a505-afc075ae420b.png)

Issue #1601, opened 2022-03-30
slcli error: 'CommandLoader' object has no attribute 'resultcallback'


hasan4791:
slcli throws error in python 3.8 & python3.9, its running on the container image redhat-ubi8.5.
Tried testing the same with python3.6 which works fine. (logs are attached)
```
root@create-cluster-testing ]# python3
Python 3.8.8 (default, Aug 11 2021, 06:52:42) 
[GCC 8.5.0 20210514 (Red Hat 8.5.0-3)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> exit()
[root@create-cluster-testing ]# 
[root@create-cluster-testing ]# slcli
bash: slcli: command not found
[root@create-cluster-testing ]# 
[root@create-cluster-testing ]# 
[root@create-cluster-testing ]# pip3 install softlayer
WARNING: Running pip install with root privileges is generally not a good idea. Try `pip3 install --user` instead.
Collecting softlayer
  Downloading https://files.pythonhosted.org/packages/ea/58/fd476cc96c19fa980ef7c1d5aaa31a35dd396f97aed34040f3cf9e5c2a5a/SoftLayer-6.0.1-py2.py3-none-any.whl (652kB)
     |████████████████████████████████| 655kB 16.9MB/s 
Collecting requests>=2.20.0
  Downloading https://files.pythonhosted.org/packages/2d/61/08076519c80041bc0ffa1a8af0cbd3bf3e2b62af10435d269a9d0f40564d/requests-2.27.1-py2.py3-none-any.whl (63kB)
     |████████████████████████████████| 71kB 15.9MB/s 
Collecting prompt-toolkit>=2
  Downloading https://files.pythonhosted.org/packages/00/e4/beb2fad0bec554a011b321d0f6b99981f9171c0b05d03a6948e45ac8a5be/prompt_toolkit-3.0.28-py3-none-any.whl (380kB)
     |████████████████████████████████| 389kB 50.7MB/s 
Collecting click>=7
  Downloading https://files.pythonhosted.org/packages/86/3e/3a523bdd24510288b1b850428e01172116a29268378b1da9a8d0b894a115/click-8.1.0-py3-none-any.whl (96kB)
     |████████████████████████████████| 102kB 16.9MB/s 
Collecting pygments>=2.0.0
[...]

PR #1605, opened 2022-03-30
slcli licenses is missing the help text

caberos:
https://github.com/softlayer/softlayer-python/issues/1604


allmightyspiff:
@caberos looks like those test errors are related to #1601 ... I'm looking into that.

allmightyspiff (approved):

Issue #1612, opened 2022-04-06
Update global ip assign/unassign to use new API


allmightyspiff:
[SoftLayer_Network_Subnet_IpAddress_Global::route](https://sldn.softlayer.com/reference/services/SoftLayer_Network_Subnet_IpAddress_Global/route/) (and unroute) is the old API

```
$ slcli -v  globalip assign 12345 192.168.1.1
Calling: SoftLayer_Network_Subnet_IpAddress_Global::route(id=12345, mask='', filter='None', args=('192.168.1.1',), limit=None, offset=None))
SoftLayerAPIError(404): Unable to find object with id of '12345'.
```

Need to change this to use https://sldn.softlayer.com/reference/services/SoftLayer_Network_Subnet/route/ instead. 
This will require using the `Network_Subnet` id for the Global IP

```
$ slcli --format=json call-api SoftLayer_Network_Subnet_IpAddress_Global getObject --id=163542 --mask="mask[id,ipAddress[id,subnet[id,cidr,gateway,subnetType]]]"
{
    "id": 163542,
    "ipAddress": {
        "id": 13082486,
        "subnet": {
            "cidr": 32,
            "gateway": "",
            "id": 524326
        }
    }
}
```

I'm not sure offhand about the exact syntax for the new Route command, but the docs are pretty verbose on it, so hopefully that will help.

PR #1614, opened 2022-04-11
Update global ip assign/unassign to use new API

caberos:
https://github.com/softlayer/softlayer-python/issues/1612


SoftLayer/CLI/globalip/assign.py
```diff
@@ -5,17 +5,17 @@
 
 import SoftLayer
 from SoftLayer.CLI import environment
-from SoftLayer.CLI import helpers
 
 
-@click.command()
+@click.command(epilog="More information about types and ")
```
allmightyspiff, line 10: Did you mean to add this text? Would be a good place to mention that the user should read https://sldn.softlayer.com/reference/services/SoftLayer_Network_Subnet/route/

SoftLayer/CLI/globalip/assign.py
```diff
@@ -5,17 +5,17 @@
...
 @click.argument('identifier')
-@click.argument('target')
+@click.option('--target',
+              help='See SLDN docs. '
+                   'E.g SoftLayer_Network_Subnet_IpAddress, SoftLayer_Hardware_Server,SoftLayer_Virtual_Guest')
```
allmightyspiff, line 14: Lets change this to a [click.choice](https://click.palletsprojects.com/en/8.0.x/options/#choice-options) option, and list out all the types in a short-hand format

so something like `type=click.Choice['vlan', 'ip', 'hardware', 'vsi' ]` Then the  network manager function can figure out which SoftLayer class needs to be sent in.

Both this and `--target-id` should be marked as required as well.

SoftLayer/CLI/globalip/assign.py
```diff
@@ -5,17 +5,17 @@
...
+@click.option('--router', help='An appropriate identifier for the specified $type. Some types have multiple identifier')
```
allmightyspiff, line 15: I think calling this option `router` might be confusing, 
```suggestion
@click.option('--target-id', help='The identifier for the destination resource to route this subnet to. ')
```

allmightyspiff (changes requested):

allmightyspiff (approved):

Issue #1622, opened 2022-04-22
Unable to get VSI details when last TXN is "Software install is finished"


andrew-edgar:
Unable to get VSI details.

➜  ~ slcli virtual detail 61354767
An unexpected error has occured:
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/SoftLayer-5.9.8-py3.9.egg/SoftLayer/CLI/core.py", line 209, in main
    cli.main(**kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/decorators.py", line 73, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/SoftLayer-5.9.8-py3.9.egg/SoftLayer/CLI/virt/detail.py", line 72, in cli
    last_transaction = "{} ({})".format(utils.lookup(result, 'lastTransaction', 'transactionGroup', 'name'),
  File "/usr/local/lib/python3.9/site-packages/SoftLayer-5.9.8-py3.9.egg/SoftLayer/utils.py", line 35, in lookup
    return lookup(dic.get(key, {}), keys[0], *keys[1:])
  File "/usr/local/lib/python3.9/site-packages/SoftLayer-5.9.8-py3.9.egg/SoftLayer/utils.py", line 35, in lookup
    return lookup(dic.get(key, {}), keys[0], *keys[1:])
AttributeError: 'list' object has no attribute 'get'

Feel free to report this error as it is likely a bug:
    https://github.com/softlayer/softlayer-python/issues
The following snippet should be able to reproduce the error

[...]

PR #1625, opened 2022-04-26
Unable to get VSI details when last TXN is "Software install is finis…

caberos:
[#1622](https://github.com/softlayer/softlayer-python/issues/1622)

### DIFF
--- a/SoftLayer/CLI/virt/detail.py
+++ b/SoftLayer/CLI/virt/detail.py
@@ -70,7 +70,7 @@ def cli(env, identifier, passwords=False, price=False):
     table.add_row(['created', result['createDate']])
     table.add_row(['modified', result['modifyDate']])
     last_transaction = ''
-    if result.get('lastTransaction') != []:
+    if result.get('lastTransaction'):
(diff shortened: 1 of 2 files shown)


SoftLayer/CLI/virt/detail.py
```diff
@@ -69,11 +69,13 @@ def cli(env, identifier, passwords=False, price=False):
     table.add_row(['transient', result.get('transientGuestFlag', False)])
     table.add_row(['created', result['createDate']])
     table.add_row(['modified', result['modifyDate']])
-    last_transaction = "{} ({})".format(utils.lookup(result, 'lastTransaction', 'transactionGroup', 'name'),
-                                        utils.clean_time(utils.lookup(result, 'lastTransaction', 'modifyDate')))
+    last_transaction = ''
+    if result.get('lastTransaction') != []:
```
allmightyspiff, line 73: You can just do the following, which will also work if `lastTransaction` is None.  Empty list and dictionaries return `False` in python.

```suggestion
    if result.get('lastTransaction'):
```

allmightyspiff (changes requested):
Thanks for taking care of this, just a minor  style change needed.

Issue #1627, opened 2022-04-28
slcli autoscale delete


caberos:
add the new command on autoscale delete

PR #1628, opened 2022-04-28
new command on autoscale delete

caberos:
[#1627](https://github.com/softlayer/softlayer-python/issues/1627)

**command**
`slcli autoscale delete scalegroupId`

### DIFF
--- a/SoftLayer/CLI/autoscale/delete.py
+++ b/SoftLayer/CLI/autoscale/delete.py
@@ -1,7 +1,7 @@
 """Delete autoscale."""
 # :license: MIT, see LICENSE for more details.
-
 import click
+import SoftLayer
 
 from SoftLayer.CLI import environment
 from SoftLayer.managers.autoscale import AutoScaleManager
@@ -18,7 +18,9 @@ def cli(env, identifier):
     """
 
     autoscale = AutoScaleManager(env.client)
-    result = autoscale.delete(identifier)
 
-    if result:
+    try:
+        autoscale.delete(identifier)
         click.secho("%s deleted successfully" % identifier, fg='green')
+    except SoftLayer.SoftLayerAPIError as ex:
+        click.secho("Failed to delete %s\n%s" % (identifier, ex), fg='red')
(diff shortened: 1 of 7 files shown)


edsonarios:
It may be necessary to ask for a confirmation to remove it, for example on dedicatedhost or vlan
image

When we send a letter, it only returns an exception, but it should return a better error message as in the image or virtual commands, but the ideal would be to specify that the parameter be an integer
image
image

SoftLayer/CLI/autoscale/delete.py
```diff
@@ -0,0 +1,28 @@
...
+
+@click.command()
+@click.argument('identifier')
+@environment.pass_env
+def cli(env, identifier):
+    """Sets a user's status to CANCEL_PENDING, which will immediately disable the account,
+
+    and will eventually be fully removed from the account by an automated internal process.
+
+    Example: slcli user delete userId
```
BrianSantivanez, line 18: Update the command description and change example by `slcli autoscale delete autoscaleId`

SoftLayer/CLI/autoscale/delete.py
```diff
@@ -0,0 +1,28 @@
...
+
+    """
+
+    autoscale = AutoScaleManager(env.client)
+    result = autoscale.delete(identifier)
+
+    if result:
+        click.secho("%s deleted successfully" % identifier, fg='green')
+    else:
```
edsonarios, line 27: This code does not seem to be necessary, because the values it returns are always "True or exception", but not false, maybe it is not necessary anymore

allmightyspiff: My recommendation here would be something like this:


```python
    try:
        result = autoscale.delete(identifier)
        click.secho("%s deleted successfully" % identifier, fg='green')
    except SoftLayer.SoftLayerAPIError as ex
        click.secho("Failed to delete %s\n%s" % (identifier, ex), fg='red')

```

allmightyspiff (changes requested):
Just need to change the message output from if/else to a try/catch and this should be good. Also make sure to update the documentation for `cli` here, as it is copied from the user delete command. Thanks

allmightyspiff (approved):

Issue #1632, opened 2022-05-05
Error on loadbal order command


caberos:
slcli loadbal order --name test --datacenter ams01 --label labeltest --frontend TCP:80 --backend TCP:80 --subnet 759282 --backend HTTP:80

SoftLayerAPIError(SoftLayer_Exception_Order_InvalidLocation): The location provided for this order is invalid. This offering is not available in ams01.

**Version**
slcli --version
Current: slcli (SoftLayer Command-line) v6.0.2

PR #1633, opened 2022-05-05
fix issue on loadbal order command

caberos:
[#1632](https://github.com/softlayer/softlayer-python/issues/1632)

### DIFF
--- a/SoftLayer/CLI/loadbal/order.py
+++ b/SoftLayer/CLI/loadbal/order.py
@@ -41,8 +41,10 @@ def order(env, **args):
     """Creates a LB. Protocols supported are TCP, HTTP, and HTTPS."""
 
     mgr = SoftLayer.LoadBalancerManager(env.client)
+    network = SoftLayer.NetworkManager(env.client)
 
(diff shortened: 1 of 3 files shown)


SoftLayer/CLI/loadbal/order.py
```diff
@@ -41,8 +41,10 @@ def order(env, **args):
     """Creates a LB. Protocols supported are TCP, HTTP, and HTTPS."""
 
     mgr = SoftLayer.LoadBalancerManager(env.client)
+    network = SoftLayer.NetworkManager(env.client)
 
-    location = args.get('datacenter')
+    datacenter = network.get_datacenter(args.get('datacenter'))
+    location = {'id': datacenter[0]['id']}
```
allmightyspiff, line 47: Add a check here to make sure we got a datacenter back, and produce an error if we didn't letting the user know the DC they want doesn't exist

allmightyspiff (changes requested):

allmightyspiff (approved):

Issue #1634, opened 2022-05-05
Policy is not added when an AutoScale Group is created 


BrianSantivanez:
> Reminder: No username or APIkeys should be added to these issues, as they are public.

**Describe the bug**
When `slcli autoscale create --name policy2 --cooldown 3600 --min 1 --max 2 -o CENTOS_7_64 --datacenter dal10 --termination-policy 2 -H testvs -D test.com -- cpu 2 --memory 1024 --policy-relative ABSOLUTE --policy-name policytest --policy-amount 3 --regional 102 --disk 25 --disk 30` is called to create an autoscale with an policy, the policy is not added

**Expected behavior**
The autoscale created should contains the policy added and it should be displayed in the UI 

**Screenshots**
![image](https://user-images.githubusercontent.com/78806346/167030031-7a8ff0ce-fc45-41c0-bda4-c079b74a4040.png)
![image](https://user-images.githubusercontent.com/78806346/167030270-9cc694b5-26d8-41f0-a168-701bd519acaa.png)

**Version**
```
slcli --version
Current: slcli (SoftLayer Command-line) v6.0.2                                                                                   
Latest:  slcli (SoftLayer Command-line) v6.0.2
```

PR #1637, opened 2022-05-06
Policy is not added when an AutoScale Group is created

caberos:
[#1634](https://github.com/softlayer/softlayer-python/issues/1634)


BrianSantivanez (approved):
Running the command
image

In the UI appears the new policy
image


Looks good to me

allmightyspiff (approved):

PR #1646, opened 2022-05-18
Rich tables

allmightyspiff:
Adds colors to tables, and basically re-writes how the table stuff is handled on the backend. 

Default Table colors
image

JSON Colors
image

Colored Debug Output:
image

## TODO

Still need to get exceptions piped through Rich Text
image

Get the unit tests to run on py3.10, not 3.1...


allmightyspiff:
## TESTING
For testing this feature, mostly make sure any commands run produce very similar output to base `slcli`.  There should also not be any Rich tags (like `[red] sdfsdfsdf [/red]` for example. Or anything like `<rich.Table.Table at  <#FFasdasdasd>`

edsonarios (approved):
Looks good to me

BrianSantivanez:
When the table has a title, the row of the title is printed without format

image

image

allmightyspiff:
@BrianSantivanez Those Titles are just the new style, its expected they won't have the border around them. 
@edsonarios About the git-bash terminal, if you are talking about the red id not being colored, I think thats fine. Your version of git-bash might not support that specific color, I had to update my version to the latest to get full color support.

Issue #1643, opened 2022-05-17
`slcli firewall detail` do not find standard firewall to display its info


BrianSantivanez:
> Reminder: No username or APIkeys should be added to these issues, as they are public.

**Describe the bug**
When `slcli firewall detail` command is set with an standard (vs or server) firewall that exists the command display a message saying that the firewall was not found

**Expected behavior**
The command should display the firewall rules

**Screenshots**
![image](https://user-images.githubusercontent.com/78806346/168892172-379d3077-9152-462e-b799-51aca89e6ec4.png)

**Version**
![image](https://user-images.githubusercontent.com/78806346/168892239-21c88880-bcab-4235-a322-d4666c33f8e2.png)

allmightyspiff:
I think the firewall command might need a complete overhaul, as the firewall stuff has changed a LOT since the original command was written.

Ideally our command should work similar to how the portal displays information https://cloud.ibm.com/netsec/firewalls#multiVlan

Mutli Vlan and Single Server firewalls are still available, but single vlan firewalls are no longer a product (although it exists since some customers still have them). We do not need to support single vlan firewalls going forward though, just multi-vlan, and single server.

caberos:
update the firewall list command to same to Ibmcloud page and add the new multi Vlan table with all data information to return this
![Uploading image.png…]()

PR #1649, opened 2022-05-24
update the firewall list

caberos:
[#issue1643](https://github.com/softlayer/softlayer-python/issues/1643)

image

### DIFF
--- a/SoftLayer/CLI/firewall/list.py
+++ b/SoftLayer/CLI/firewall/list.py
@@ -88,7 +88,7 @@ def cli(env):
                                  gatewalls['networkFirewall']['datacenter']['name'],
                                  gatewalls['publicIpAddress']['ipAddress'],
                                  gatewalls['privateIpAddress']['ipAddress'],
-                                 len(gatewalls['insideVlans']), gatewalls['status']['keyName']]),
+                                 len(gatewalls['insideVlans']), gatewalls['status']['keyName']])
     env.fout(table)
     env.fout(table_gatewalls)
 
--- a/SoftLayer/CLI/firewall/list.py
+++ b/SoftLayer/CLI/firewall/list.py
@@ -18,7 +18,7 @@ def cli(env):
     table = formatting.Table(['firewall id',
                               'type',
                               'features',
-                              'server/vlan id'])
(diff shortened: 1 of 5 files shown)


SoftLayer/CLI/firewall/list.py
```diff
@@ -70,7 +70,27 @@ def cli(env):
...
+                                        'Hostname',
+                                        'Location',
+                                        'Public Ip',
+                                        'Private Ip',
+                                        'Associated vlan',
+                                        'status'], title='Multi Vlan Firewall')
+    fw_gatewwalls = mgr.get_firewalls_gatewalls()
+
+    for gatewalls in fw_gatewwalls:
+        table_gatewalls.add_row([gatewalls.get('id'), gatewalls.get('name'),
```
BrianSantivanez, line 85: Display the ID with the format `multiVlan:FIREWALL_ID`, where `multiVlan` is constant and `FIREWALL_ID` is the firewall id

SoftLayer/managers/firewall.py
```diff
@@ -302,3 +302,19 @@ def get_instance(self, firewall_id, mask=None):
         svc = self.client['Network_Vlan_Firewall']
 
         return svc.getObject(id=firewall_id, mask=mask)
+
+    def get_firewalls_gatewalls(self):
+        """Returns a list of all firewalls on the account.
+
+        returns: A list of firewalls on the current account.
+        """
```
allmightyspiff, line 310: I dont mean to nit-pick here, but `gatewall` isn't an English word. However as a mashup of "gateway" + "firewall" I do kinda like it, and want to keep it. Just make the following change to the docblock so anyone else knows you put `gatewalls` on purpose, and just didn't misspell `gateways`, which is what I thought when I first saw it lol.

```suggestion
    def get_firewalls_gatewalls(self):
        """Returns a list of all gateway firewalls (gatewalls) on the account.

        returns: A list of gateway firewalls (gatewalls) on the current account.
        """
```

allmightyspiff (changes requested):
Looks good, just a minor doc change needed please.

Thanks

allmightyspiff (approved):

Issue #1654, opened 2022-05-26
add firewall monitoring


caberos:
simulate as https://cloud.ibm.com/netsec/firewalls/multi-vlan/17438/details#monitoring

![image](https://user-images.githubusercontent.com/22535437/170587848-0fcb4f36-d982-4fc8-abfe-7011600df871.png)

PR #1657, opened 2022-06-02
add firewall monitoring command

caberos:
[#issue1654](https://github.com/softlayer/softlayer-python/issues/1654)

slcli firewall monitoring firewallId

┌────────────┬─────────────────────┐
│       name │ value               │
├────────────┼─────────────────────┤
│         Id │ 17438               │
│       Name │ testFirewall-cgallo │
│ Stard Date │ 2022-05-02          │
│   End Date │ 2022-06-01          │
│        Out │ 62852476.0          │
│         In │ 89715092.0          │
│      Total │ 152567568.0         │
│     Status │ Active              │
└────────────┴─────────────────────┘


BrianSantivanez:
Running command with multi vlan firewall id 
image
Running command with single server firewall id
image
Running help command 
image

The command should display an error message saying that single server firewalls  does not have monitors or in other case update help command to say that command only accepts multi vlan firewall Ids

caberos:
only accepts multi vlan firewall because it has a metric tracking object for show the usage and monitoring the its data(input and output information), I will update the description and requeriments to use in this command
image

SoftLayer/CLI/firewall/monitoring.py
```diff
@@ -0,0 +1,51 @@
...
+        if monitor['type'] == 'publicIn_net_octet':
+            public_in += monitor['counter']
+
+    table.add_row(['Id', _firewall.get('id')])
+    table.add_row(['Name', _firewall['networkGateway']['name']])
+    table.add_row(['Stard Date', last_month.strftime('%Y-%m-%d')])
+    table.add_row(['End Date', end.strftime('%Y-%m-%d')])
+    table.add_row(['Out', public_out])
+    table.add_row(['In', public_in])
+    table.add_row(['Total', public_out + public_in])
```
allmightyspiff, line 48: Lets display this in GB using https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/CLI/formatting.py#L88

SoftLayer/managers/firewall.py
```diff
@@ -319,3 +320,20 @@ def get_firewalls_gatewalls(self):
...
+            "keyName": "PUBLICIN",
+            "summaryType": "sum"
+
+        }, {
+            "keyName": "PUBLICOUT",
+            "summaryType": "sum"
+        }]
+
+        return self.client['Metric_Tracking_Object'].getSummaryData(start_date,
+                                                                    end_date, body, 1800, id=identifier)
```
allmightyspiff, line 339: Since were getting data for a whole month, lets use the max summary period so we get less data from the API
```suggestion
        return self.client['Metric_Tracking_Object'].getSummaryData(start_date,
                                                                    end_date, body, 86400 , id=identifier)
```

SoftLayer/CLI/firewall/monitoring.py
```diff
@@ -0,0 +1,51 @@
...
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+
+
+@click.command(cls=SoftLayer.CLI.command.SLCommand, )
+@click.argument('identifier')
+@environment.pass_env
+def cli(env, identifier):
+    """Get details for a firewall monitors device."""
```
allmightyspiff: Update this description to be more specific of what it is actually doing
```suggestion
    """Gets bandwidth details for a firewall from the past 30 days"""
```

Allowing for a custom date range would be useful, but that can be added in a future issue

SoftLayer/CLI/firewall/monitoring.py
```diff
@@ -0,0 +1,51 @@
+"""Get monitoring for a firewall device."""
```
allmightyspiff, line 1:
```suggestion
"""Gets bandwidth details for a firewall from the past 30 days"""
```

allmightyspiff (changes requested):
Code looks good, just a few minor changes. Also I'd like this command to be called 'slcli firewall bandwidth` as that would align better with the hw and vs bandwidth commands.

For a `monitoring` command, I would expect it to tell me if the system is UP/DOWN and things like that.

Thanks

allmightyspiff (approved):

Issue #1683, opened 2022-07-05
add the file volume-options command


caberos:
![image](https://user-images.githubusercontent.com/22535437/177339175-80679f89-bb0e-4f80-9bb1-c043e62af5bc.png)

**slcli version**
slcli (SoftLayer Command-line) v6.1.0

PR #1684, opened 2022-07-06
add the file volume-options command

caberos:
[#issue1683](https://github.com/softlayer/softlayer-python/issues/1683)

image


allmightyspiff (changes requested):
Same comments as the `block volume-options` command.

Issue #1616, opened 2022-04-18
Improve responses of the commands


edsonarios:
There are commands that do not have an adequate response or sometimes have no response at all, so these responses need to be improved

| Command| Done? | PR |
| --- | --- | --- |
| account cancel-item | ⌛ | #1617 |

PR #1695, opened 2022-07-13
Improved successful response to command - slcli vs cancel

edsonarios:
Main Issue: #1616 
```
softlayer-python/ (issue1616-virtual-cancel) $ slcli vs cancel 123456
This action cannot be undone! Type "123456" or press Enter to abort: 123456
The virtual server instance: 123456 was cancelled.
```


BrianSantivanez (approved):
image

Looks good to me

allmightyspiff (approved):

Issue #1708, opened 2022-07-26
Command slcli vlan edit accept that we do not send any parameters


edsonarios:
The command allows not send any parameter and still displays a message that it was edited correctly, that also affects when an invalid id is sent

![image](https://user-images.githubusercontent.com/43519769/181035508-67bca543-f9fe-468c-a09d-82e30c11e5b2.png)

```
~/ $ slcli --version
Current: slcli (SoftLayer Command-line) v6.1.0
Latest:  slcli (SoftLayer Command-line) v6.1.0
```

PR #1709, opened 2022-07-26
Command slcli vlan edit accept that we do not send any parameters

edsonarios:
Issue link #1708
```
softlayer-python/ (issue1708) $ slcli vlan edit 12345
Is required Name or Note or Tags
```

### DIFF
--- a/SoftLayer/CLI/vlan/edit.py
+++ b/SoftLayer/CLI/vlan/edit.py
@@ -5,6 +5,7 @@ import click
 
 import SoftLayer
 from SoftLayer.CLI import environment
+from SoftLayer.CLI import exceptions
 from SoftLayer.CLI import helpers
 
 
@@ -21,6 +22,9 @@ from SoftLayer.CLI import helpers
 @environment.pass_env
 def cli(env, identifier, name, note, tags):
     """Edit a vlan's details."""
+
+    if not (name or note or tags):
+        raise exceptions.CLIAbort("Is required Name or Note or Tags")
 
     new_tags = None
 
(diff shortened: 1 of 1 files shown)


caberos (approved):
LGTM

SoftLayer/CLI/vlan/edit.py
```diff
@@ -22,6 +23,9 @@
 def cli(env, identifier, name, note, tags):
     """Edit a vlan's details."""
 
+    if not (name or note or tags):
```
allmightyspiff, line 26: For more than 2 options, I prefer using [`any()`](https://docs.python.org/3/library/functions.html#any)  as I feel it is a little easier to read.
```suggestion
    if not any([name, note, tags]):
```

SoftLayer/CLI/vlan/edit.py
```diff
@@ -22,6 +23,9 @@
...
+        raise exceptions.CLIAbort("Is required Name or Note or Tags")
```
allmightyspiff, line 27: "Name, Note or Tags is required" would be a more natural sounding sentence. But for this case, I'd prefer the following.

```suggestion
        raise exceptions.CLIAbort("At least one option is required")
```

allmightyspiff (changes requested):
Just some minor style changes needed, thanks.

SoftLayer/CLI/vlan/edit.py
edsonarios: Done

SoftLayer/CLI/vlan/edit.py
edsonarios: Done

allmightyspiff (approved):

Issue #1749, opened 2022-09-07
Unhandled error running a subcommand in slcli


edsonarios:
```
~/ $ slcli virtual capacity
Usage: slcli virtual capacity [OPTIONS] COMMAND [ARGS]...

  Base command for all capacity related concerns

Options:
  -h, --help  Show this message and exit.

Commands:
  create          Create a Reserved Capacity instance.
  create-guest    Allows for creating a virtual guest in a reserved capacity.
  create-options  List options for creating Reserved Capacity
  detail          Reserved Capacity Group details.
  list            List Reserved Capacity groups.
```
```
~/ $ slcli virtual capacity abcd
An unexpected error has occured:
Traceback (most recent call last):
  File "\AppData\Local\Programs\Python\Python310\lib\site-packages\SoftLayer\CLI\core.py", line 182, in main
    cli.main(**kwargs)
  File "\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1653, in invoke
    cmd_name, cmd, args = self.resolve_command(ctx, args)
  File "\AppData\Local\Programs\Python\Python310\lib\site-packages\click\core.py", line 1700, in resolve_command
    cmd = self.get_command(ctx, cmd_name)
  File "\AppData\Local\Programs\Python\Python310\lib\site-packages\SoftLayer\CLI\virt\capacity\__init__.py", line 40, in get_command
    module = importlib.import_module(path)
  File "\AppData\Local\Programs\Python\Python310\lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'SoftLayer.CLI.virt.capacity.abcd'

[...]

PR #1754, opened 2022-09-21
Unhandled error running a subcommand in slcli

caberos:
[#issue1749](https://github.com/softlayer/softlayer-python/issues/1749)

image


allmightyspiff (changes requested):
You should just be able to add `return None` to your except statements to fix the `analysis` error. Otherwise looks good.

allmightyspiff (approved):

Issue #1773, opened 2022-10-19
add hardware create credentials command


caberos:
simulate this page 
https://cloud.ibm.com/gen1/infrastructure/bare-metal/1532729/details#passwords

![image](https://user-images.githubusercontent.com/22535437/196787796-fcb89375-5873-4927-a0e8-f83be9847b5b.png)

slcli version: 6.1.2

PR #1774, opened 2022-10-21
new command hw create-credential

caberos:
[#issue1677](https://github.com/softlayer/softlayer-python/issues/1773)

image

### DIFF
--- a/SoftLayer/CLI/hardware/create_credential.py
+++ b/SoftLayer/CLI/hardware/create_credential.py
@@ -22,14 +22,13 @@ def cli(env, identifier, username, password, notes, system):
 
     software = mgr.get_software_components(identifier)
     sw_id = None
-    for sw in software:
-        if (sw['softwareLicense']['softwareDescription']['name']).lower() == system:
-            sw_id = sw['id']
+    for sw_instance in software:
+        if (sw_instance['softwareLicense']['softwareDescription']['name']).lower() == system:
+            sw_id = sw_instance['id']
             break
         elif system:
-            if (sw['softwareLicense']['softwareDescription']['name']) == 'Passmark Suite':
-                sw_id = sw['id']
-                break
+            if (sw_instance['softwareLicense']['softwareDescription']['name']) == 'Passmark Suite':
+                sw_id = sw_instance['id']
 
     template = {
         "notes": notes,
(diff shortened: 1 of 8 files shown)


SoftLayer/CLI/hardware/create_credential.py
```diff
@@ -0,0 +1,57 @@
...
+
+    software = mgr.get_software_components(identifier)
+    sw_id = None
+    for sw_instance in software:
+        if (sw_instance['softwareLicense']['softwareDescription']['name']).lower() == system:
+            sw_id = sw_instance['id']
+            break
+        elif system:
+            if (sw_instance['softwareLicense']['softwareDescription']['name']) == 'Passmark Suite':
+                sw_id = sw_instance['id']
```
allmightyspiff, line 31: I dont understand what you are checking for here?

Looks to me like you are checking for if a user picked `--system`, but then you just set sw_id to `Passmark Suite`. I feel like this whole `elif` check could be removed, and if at the end if your for loop, if `sw_id` is None and `system` is not None, throw and exception about not being able to find a software called `system`

SoftLayer/managers/hardware.py
```diff
@@ -1102,6 +1102,14 @@ def add_notification(self, hardware_id, user_id):
         template = {"hardwareId": hardware_id, "userId": user_id}
         return self.client.call('SoftLayer_User_Customer_Notification_Hardware', 'createObject', template)
 
+    def get_software_components(self, hardware_id):
+        """Returns Hardware sensor data"""
```
allmightyspiff, line 1106: Fix this doc line to reflect what this function actually does.

SoftLayer/CLI/hardware/create_credential.py
```diff
@@ -0,0 +1,57 @@
...
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+
+
+@click.command(cls=SoftLayer.CLI.command.SLCommand, )
+@click.argument('identifier')
+@click.option('--username', '-U', help="The username part of the username/password pair")
+@click.option('--password', '-P', help="The password part of the username/password pair.")
+@click.option('--notes', '-n', help="A note string stored for this username/password pair.")
+@click.option('--system', help="The name of this specific piece of software.")
```
allmightyspiff, line 16: If these are required, you should mark them as such, and if not, set a default value.

allmightyspiff (changes requested):

edsonarios (approved):
Looks good to me

allmightyspiff (approved):

Issue #1652, opened 2022-05-25
Add a 'light' theme to the slcli


allmightyspiff:
Add a theme that looks good on white backgrounds, and a config option to set it

PR #1775, opened 2022-10-21
New feature to change theme in slcli like dark, light o maintain in default

edsonarios:
Issue link: #1652
New feature to change theme in slcli like dark, light o maintain in default

With this new feature you can choose the theme for slcli like "dark, light" or if you not set theme, will use default theme
You can see the theme with the command `slcli config show`
image

I'ts possible change to theme with the follow command `slcli config setup`
image

Also, it's possible to set the env variable with `export SL_THEME`

Example
image
image

For now the color changes are small because they only adapt to see better according to the background, but now you can add new themes, for more drastic color changes

### DIFF
--- a/SoftLayer/utils.py
+++ b/SoftLayer/utils.py
@@ -11,6 +11,8 @@ from json import JSONDecoder
 import re
 import time
 
+from rich.console import Console
+from rich.theme import Theme
 
 # pylint: disable=no-member, invalid-name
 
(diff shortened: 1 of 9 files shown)


SoftLayer/utils.py
```diff
@@ -459,3 +461,74 @@ def decode_stacked(document, pos=0, decoder=JSONDecoder()):
...
+                "name_sub_command": "orange3",  # sub command name
+                "sub_command": "orange3",  # sub command list
+                # Help table colors options
+                "option": "bold cyan",
+                "switch": "bold green",
+                "default_option": "light_coral",
+                "option_keyword": "bold cyan",
+                "args_keyword": "bold green",
+            })
+        )
```
allmightyspiff, line 485: We dont need to re-define the dark theme here.

basically just refactor this to be 

```python

if theme == 'light':
    # return the light theme console
else:
    # return the dark theme console
```

SoftLayer/utils.py
```diff
@@ -459,3 +461,74 @@ def decode_stacked(document, pos=0, decoder=JSONDecoder()):
...
+    return None
+
+
+def table_color_theme(theme):
+    """Define result table colors"""
+    if not theme:
+        return {'header': 'bright_cyan',
+                'id_columns': 'pale_violet_red1'}
```
allmightyspiff, line 527: Same comment as above, do 

```python
if theme == 'light':
    # light theme
else:
    # dark theme
```

SoftLayer/utils.py
```diff
@@ -459,3 +461,74 @@ def decode_stacked(document, pos=0, decoder=JSONDecoder()):
         obj, pos = decoder.raw_decode(document, pos)
 
         yield obj
+
+
+def console_color_themes(theme):
```
allmightyspiff, line 466: These functions might be better in the Environment file, I think that will be the only place that uses them (with the changes I made in https://github.com/edsonarios/softlayer-python/pull/3/files)

allmightyspiff (changes requested):

SoftLayer/utils.py
edsonarios: Done

SoftLayer/utils.py
edsonarios: Done

allmightyspiff:
@caberos do you mind testing this out since I think you mostly use a light theme.

allmightyspiff (approved):

caberos (approved):
LGTM

Issue #1771, opened 2022-10-17
slcli hw add-notifications crashes with a bad user


allmightyspiff:
```
$ ./slcli -vvv  hw add-notification  1721429 --users 12345
Calling: SoftLayer_User_Customer_Notification_Hardware::createObject(id=None, mask='', filter='None', args=({'hardwareId': '1721429', 'userId': '12345'},), limit=None, offset=None))
Starting new HTTPS connection (1): api.softlayer.com:443
https://api.softlayer.com:443 "POST /rest/v3.1/SoftLayer_User_Customer_Notification_Hardware/createObject.json HTTP/1.1" 200 62
Returned Data:
{'error': 'Internal Error', 'code': 'SoftLayer_Exception_Public'}
An unexpected error has occured:

Traceback (most recent call last):
  File "C:\Users\allmi\Source\softlayer-python\SoftLayer\CLI\core.py", line 182, in main
    cli.main(**kwargs)
  File "C:\Users\allmi\Source\py38\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "C:\Users\allmi\Source\py38\lib\site-packages\click\core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Users\allmi\Source\py38\lib\site-packages\click\core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Users\allmi\Source\py38\lib\site-packages\click\core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\allmi\Source\py38\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "C:\Users\allmi\Source\py38\lib\site-packages\click\decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "C:\Users\allmi\Source\py38\lib\site-packages\click\core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "C:\Users\allmi\Source\softlayer-python\SoftLayer\CLI\hardware\add_notification.py", line 27, in cli
    table.add_row([notification['id'], notification['hardware']['fullyQualifiedDomainName'],
KeyError: 'id'

Feel free to report this error as it is likely a bug:
    https://github.com/softlayer/softlayer-python/issues
The following snippet should be able to reproduce the error
```

[...]

PR #1786, opened 2022-11-01
slcli hw add-notifications crashes with a bad user

caberos:
[#issue177](https://github.com/softlayer/softlayer-python/issues/1771)

image


BrianSantivanez (approved):
image
Looks good to me

edsonarios:
When introducing an ID invalid just show an empty table, I thinks, should show us an error
image

allmightyspiff (changes requested):
Looks good, however I think we should make `--users` a required option, which would also address Edson's comments.

allmightyspiff (approved):

Issue #1768, opened 2022-10-13
Improve firewall cancel command


allmightyspiff:
I can't seem to cancel this multiVlan firewall. I should be able to cancel it. The help message should also show the available types of firewalls. Also add that information to the detail command
```
$ slcli firewall cancel multiVlan:17438
This action will cancel a firewall from your account. Continue? [y/N]: y
Unknown firewall type: multiVlan

$ slcli firewall cancel 17438
Invalid ID 17438: ID should be of the form xxx:yyy

$ slcli firewall cancel --help
Usage: slcli firewall cancel [OPTIONS] IDENTIFIER

        Cancels a firewall.

┌────┬────────────┬─────────────────────────────┐
│    │ identifier │                             │
│ -h │ --help     │ Show this message and exit. │
└────┴────────────┴─────────────────────────────┘


$ slcli firewall list
             Single Server Firewalls
┌─────────────┬──────┬──────────┬────────────────┐
│ firewall id │ type │ features │ server/vlan id │
├─────────────┼──────┼──────────┼────────────────┤
└─────────────┴──────┴──────────┴────────────────┘
                                                                     Multi Vlan Firewall
┌───────┬─────────────────────┬───────────────────────────────────┬───────────────────┬──────────┬────────────────┬──────────────┬─────────────────┬────────┐
│  Id   │      firewall       │               type                │     Hostname      │ Location │   Public Ip    │  Private Ip  │ Associated vlan │ status │
├───────┼─────────────────────┼───────────────────────────────────┼───────────────────┼──────────┼────────────────┼──────────────┼─────────────────┼────────┤
│ 17438 │ testFirewall-cgallo │ fortigate-security-appliance-10gb │ dft03.pod03.dal13 │  dal13   │ 67.228.206.245 │ 10.37.115.70 │        0        │ ACTIVE │
└───────┴─────────────────────┴───────────────────────────────────┴───────────────────┴──────────┴────────────────┴──────────────┴─────────────────┴────────┘
```

allmightyspiff:
## TODO
- Refactor this cancel command to we don't have to specify 'multiVlan', which might be done as part of #1820 
- Once we have the ID, we should be able to cancel it by looking up the billing item and putting in a cancellation for that.

PR #1787, opened 2022-11-02
Improve firewall cancel command

caberos:
[#issue1768](https://github.com/softlayer/softlayer-python/issues/1768)

image


edsonarios:
There is still problems
image

allmightyspiff (changes requested):
Right now there is no way to actually cancel a multiVlan firewall with this command. We either need to improve this command to allow that, or maybe make a new command if that makes sense.

allmightyspiff:
Lets revisit this topic once we refactor the firewall identifier stuff (#1820 ). basically I don't want to deal with having users specify `multiVlan` or `singleServer` when interacting with firewalls. I think once we get the Searching features in, we can revisit this topic.

Issue #1788, opened 2022-11-03
Call API can't handle empty results sometimes


allmightyspiff:
```bash
$ ./slcli -vvv  --format=json call-api SoftLayer_Product_Package_Preset getObject --id=337 --mask="mask[id,prices,configuration]"
Calling: SoftLayer_Product_Package_Preset::getObject(id=337, mask='mask[id,prices,configuration]', filter='{}', args=(), limit=None, offset=None))
Starting new HTTPS connection (1): api.softlayer.com:443
https://api.softlayer.com:443 "GET /rest/v3.1/SoftLayer_Product_Package_Preset/337/getObject.json?objectMask=mask%5Bid%2Cprices%2Cconfiguration%5D HTTP/1.1" 200 826
Returned Data:
{'id': 337, 'configuration': [{}, {}, {}], 'prices': [{'currentPriceFlag': None, 'hourlyRecurringFee': '.016', 'id': 1639, 'itemId': 865, 'laborFee': '0', 'locationGroupId': None, 'onSaleFlag': None, 'oneTimeFee': '0', 'quantity': None, 'recurringFee': '10.29', 'setupFee': '0', 'sort': 0, 'termLength': None, 'tierMinimumThreshold': None}, {'currentPriceFlag': None, 'hourlyRecurringFee': '.06', 'id': 234756, 'itemId': 13748, 'laborFee': '0', 'locationGroupId': None, 'onSaleFlag': None, 'oneTimeFee': '0', 'quantity': None, 'recurringFee': '40', 'setupFee': '0', 'sort': 0, 'termLength': None, 'tierMinimumThreshold': None}, {'currentPriceFlag': None, 'hourlyRecurringFee': '.036', 'id': 204181, 'itemId': 10567, 'laborFee': '0', 'locationGroupId': None, 'onSaleFlag': None, 'oneTimeFee': '0', 'quantity': None, 'recurringFee': '24', 'setupFee': '0', 'sort': 0, 'termLength': None, 'tierMinimumThreshold': None}]}
An unexpected error has occured:
Traceback (most recent call last):
  File "C:\Users\allmi\Source\softlayer-python\SoftLayer\CLI\core.py", line 183, in main
    cli.main(**kwargs)
  File "C:\Users\allmi\Source\py38\lib\site-packages\click\core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "C:\Users\allmi\Source\py38\lib\site-packages\click\core.py", line 1657, in invoke
[...]

PR #1789, opened 2022-11-09
fix the call api cannot handle empty results

caberos:
[#issue1788](https://github.com/softlayer/softlayer-python/issues/1788)

image

### DIFF
--- a/SoftLayer/CLI/formatting.py
+++ b/SoftLayer/CLI/formatting.py
@@ -400,18 +400,9 @@ def _format_dict(result):
     table.align['value'] = 'l'
 
     for key, value in result.items():
+        value = iter_to_table(value)
+        table.add_row([key, value])
 
-        table_list = Table(['value'])
-        if str(value) == "[[], [], []]":
-            for item in value:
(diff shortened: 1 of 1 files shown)


allmightyspiff (changes requested):
This will only resolve the issue when the API results are exactly `[[], [], []]` which isn't great.

What you should do, is add a check to https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/CLI/formatting.py#L410.

something like
```python
def _format_list(result):
    """Format list responses into a table."""
    
    if not result:
        return result
    table = Table(['value'])
    new_result = [item for item in result if item]
    if len(new_result) == 0: 
        table.add_row(["-"])
        return table
    if isinstance(new_result[0], dict):
        return _format_list_objects(new_result)

   
    for item in new_result:
        table.add_row([iter_to_table(item)])
    return table
```

And adding a unit test for this specific situation would be good too.

Thanks

allmightyspiff:
image

Looks good, thanks

allmightyspiff (approved):

Issue #1802, opened 2022-12-05
`slcli vs upgrade --resize-disk` is resizing another disk and not the selected one


BrianSantivanez:
> Reminder: No username or APIkeys should be added to these issues, as they are public.

**Describe the bug**
The command requires two arguments, the new size in GB and the disk number to resize, the command is resizing a disk but not the selected disk

**Expected behavior**
The command should resizing the selected disk

**Screenshots**
Running the command 
![image](https://user-images.githubusercontent.com/78806346/205746008-76cd3235-5652-40e0-b512-09f3f1d5df0a.png)
Before and after run the command
![image](https://user-images.githubusercontent.com/78806346/205747275-d399ca57-f90b-4264-862c-dad2b437c9a1.png)

**Version**
![image](https://user-images.githubusercontent.com/78806346/205746410-a81715e4-37a1-46e4-9ce2-6a55dfe22ed7.png)

PR #1804, opened 2022-12-06
`slcli vs upgrade --resize-disk` is resizing another disk and not the selected one

BrianSantivanez:
Issue: https://github.com/softlayer/softlayer-python/issues/1802
Observations: The command was updated to resize disk using category id instead category code
```
slcli vs upgrade 134499384 --resize-disk 10 4
This action will incur charges on your account. Continue? [y/N]: y
```
image

### DIFF
--- a/SoftLayer/managers/vs.py
+++ b/SoftLayer/managers/vs.py
@@ -1068,8 +1068,11 @@ class VSManager(utils.IdentifierMixin, object):
                                                        'Unable to find %s option with value %s' % (
                                                            ('disk', disk_guest.get('capacity'))))
 
+                category_id = self.get_disk_category_id_by_disk_number(disk_guest.get('capacity'), disk_number)
+                if category_id is None:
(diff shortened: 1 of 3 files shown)


SoftLayer/managers/vs.py
```diff
@@ -1101,6 +1104,40 @@ def upgrade(self, instance_id, cpus=None, memory=None, nic_speed=None, public=Tr
             return True
         return False
 
+    def get_disk_category_id_by_disk_number(self, capacity, disk_number):
+        """Uses Product_Package::getItemPrices to get all disk items with its categories."""
+        disk_key_names = {
```
caberos, line 1109: add information or reference why use this json

caberos (changes requested):

SoftLayer/managers/vs.py
BrianSantivanez: Done

edsonarios (approved):
Looks good to me

allmightyspiff (approved):

Issue #1806, opened 2022-12-12
add a limit and sortby on data time on ticket list command  on result delay


caberos:
add a limit with a sortby to return the last tickets(closed and opens) and filter to get by description

![image](https://user-images.githubusercontent.com/22535437/207172597-7c19e362-8787-41e8-a75a-d28845bb02bd.png)

slcli version: 6.1.3

PR #1808, opened 2022-12-13
add a limit and sortby on data time on ticket list command on result …

caberos:
[#issue1806](https://github.com/softlayer/softlayer-python/issues/1806)

image

### DIFF
--- a/SoftLayer/managers/ticket.py
+++ b/SoftLayer/managers/ticket.py
@@ -39,7 +39,7 @@ class TicketManager(utils.IdentifierMixin, object):
                 call = 'getClosedTickets'
             else:
                 raise ValueError("open_status and closed_status cannot both be False")
-        return self.client.call('Account', call, mask=mask, limit=100)
+        return self.client.call('Account', call, mask=mask, iter=False,limit=100)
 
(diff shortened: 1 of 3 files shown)


SoftLayer/managers/ticket.py
```diff
@@ -39,7 +39,7 @@ def list_tickets(self, open_status=True, closed_status=True):
                 call = 'getClosedTickets'
             else:
                 raise ValueError("open_status and closed_status cannot both be False")
-        return self.client.call('Account', call, mask=mask, iter=True)
+        return self.client.call('Account', call, mask=mask, limit=100)
```
allmightyspiff, line 42: I'd prefer we modified the `list_tickets` method to include parameters (with defaults) for `iter = False`, `limit=100`, `offset=0` so that way its easier for the CLI commands to pass in those parameters if needed.

Change looks good otherwise, thanks

allmightyspiff (changes requested):

edsonarios:
Screenshot of the test result
image

edsonarios (approved):
Looks good to me

allmightyspiff (approved):

PR #1832, opened 2023-01-18
update volume-refresh command

ko101:


allmightyspiff (changes requested):
Please move these files to their proper location.

`refresh.py` should be in `SoftLayer/CLI/file/refresh.py`
`stirage.py` should be in `SoftLayer/managers/storage.py`

allmightyspiff:
@ko101 you'll need to make changes to your [ko101:master](https://github.com/ko101/softlayer-python) branch before the changes are reflected in this pull request.

These 2 files need to be moved to their proper place in the repository.  If that doesn't make sense let me know.

SoftLayer/CLI/block/refresh.py
```diff
@@ -9,10 +9,15 @@
 @click.command(cls=SoftLayer.CLI.command.SLCommand, )
 @click.argument('volume_id')
 @click.argument('snapshot_id')
+@click.option('--force-refresh', '-f',
+    help = "Cancel current refresh process and initiates the new refresh.",
+    type = click.BOOL,
```
allmightyspiff, line 14:
```suggestion
    is_flag=True,
```

Have this option set to a flag which will allow users to do `slcli block refresh VOLUMEID SNAPSHOTID --force-refresh` instead of `slcli block refresh VOLUMEID SNAPSHOTID --force-refresh 1`

Also, these lines need to be aligned with the first `'` on line 12, which is what tox analysis is complaining about

SoftLayer/CLI/file/refresh.py
```diff
@@ -9,10 +9,15 @@
 @click.command(cls=SoftLayer.CLI.command.SLCommand, )
 @click.argument('volume_id')
 @click.argument('snapshot_id')
+@click.option('--force-refresh', '-f',
+    help = "Cancel current refresh process and initiates the new refresh.",
+    type = click.BOOL,
+    default = False,
+    show_default = True)
```
allmightyspiff, line 16: Use this style, here and in the other command.
```suggestion
@click.option('--force-refresh', '-f', is_flag = True, default = False, show_default = True,
              help = "Cancel current refresh process and initiates the new refresh."
)
```

allmightyspiff (changes requested):
Looking better, just needs some style adjustments. Thanks

SoftLayer/CLI/file/refresh.py
ko101: Sure, will look into suggestions and fix indentation as well

SoftLayer/CLI/block/refresh.py
ko101: @allmightyspiff i agree this is more simpler, but our API's dev want existing way only, Is it fine we do this way passing value along with param?

allmightyspiff: When setting `is_flag=True` `force-refresh` will be a boolean value, and it will make the same api command. This simply makes the command easier for the users.   If `--force-refresh` is present as an option, it will be True, if it is not present, it will be False.

allmightyspiff (approved):
Looks good, just waiting on the API to be available before merging in, FBLOCK-4704 should be out 2023-02-06

allmightyspiff (approved):

Issue #1837, opened 2023-01-24
os-available feature on vs


caberos:
return the all Operation system available associate to account

[#issue1826](https://github.com/softlayer/softlayer-python/issues/1826)

![image](https://user-images.githubusercontent.com/22535437/214429445-d2864cff-d3d8-45e4-bd58-b925b2393a2c.png)

PR #1835, opened 2023-01-23
New Command: `slcli virtual host-list`

BrianSantivanez:
Issue: https://github.com/softlayer/softlayer-python/issues/1829
Observations: virtual host-list command was added with its unit tests
```
slcli vs host-list --order 85857762 -n dedicatedhost01 -d dal13 --owner sl307608-dcabero
┌────────┬─────────────────┬────────────┬──────────────┬───────────────────────┬──────────────────────────┬────────────────────────┬────────┐
│   Id   │ Name            │ Datacenter │    Router    │ CPU (allocated/total) │ Memory (allocated/total) │ Disk (allocated/total) │ Guests │
├────────┼─────────────────┼────────────┼──────────────┼───────────────────────┼──────────────────────────┼────────────────────────┼────────┤
│ 656700 │ dedicatedhost01 │   dal13    │ bcr01a.dal13 │         0/56          │          0/242           │         0/1200         │   0    │
└────────┴─────────────────┴────────────┴──────────────┴───────────────────────┴──────────────────────────┴────────────────────────┴────────┘
```

PR #1838, opened 2023-01-24
vs os-available command

caberos:
[#issue1837](https://github.com/softlayer/softlayer-python/issues/1837)

image


edsonarios:
Screenshots of the test result
image

SoftLayer/CLI/virt/os_available.py
```diff
@@ -0,0 +1,28 @@
...
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+
+
+@click.command(cls=SoftLayer.CLI.command.SLCommand, )
+@environment.pass_env
+def cli(env):
+    """Get the all Os available"""
+
+    table = formatting.KeyValueTable(['id', 'keyName', 'Description', 'price', 'setup fee'])
```
edsonarios, line 16: The key values are better in UpperLower case letters
```suggestion
    table = formatting.KeyValueTable(['Id', 'KeyName', 'Description', 'Price', 'Setup Fee'])
```

SoftLayer/CLI/virt/os_available.py
```diff
@@ -0,0 +1,28 @@
+"""Get the all Os available."""
```
allmightyspiff, line 1:
```suggestion
"""Get all available Operating Systems."""
```

SoftLayer/CLI/virt/os_available.py
```diff
@@ -0,0 +1,28 @@
...
+
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+
+
+@click.command(cls=SoftLayer.CLI.command.SLCommand, )
+@environment.pass_env
+def cli(env):
+    """Get the all Os available"""
```
allmightyspiff, line 14:
```suggestion
    """Get all available Operating Systems."""
```

SoftLayer/CLI/virt/os_available.py
```diff
@@ -0,0 +1,28 @@
...
+
+    table = formatting.KeyValueTable(['Id', 'KeyName', 'Description', 'Price', 'Setup Fee'])
+    table.align['name'] = 'r'
+    table.align['value'] = 'l'
+    manager = SoftLayer.OrderingManager(env.client)
+    _filter = {"items": {"prices": {"categories": {"categoryCode": {"operation": "os"}}}}}
+    operations = manager.get_items(storage_filter=_filter, package_id=46)
```
allmightyspiff, line 21: Use the `PUBLIC_CLOUD_SERVER` package, as I think that is more inline with what is in the portal. The results are likely very similar, but I think this is the default package now.

```suggestion
    # PUBLIC_CLOUD_SERVER = 835
    operations = manager.get_items(storage_filter=_filter, package_id=835)
```

It might also be a good idea to add this package as a constant in the ordering manager. I'm OK with having the Id here as its not likely to change ever, but having a constant for it would be more "proper"  I think.

SoftLayer/CLI/virt/os_available.py
```diff
@@ -0,0 +1,28 @@
...
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import formatting
+
+
+@click.command(cls=SoftLayer.CLI.command.SLCommand, )
+@environment.pass_env
+def cli(env):
+    """Get the all Os available"""
+
+    table = formatting.KeyValueTable(['Id', 'KeyName', 'Description', 'Price', 'Setup Fee'])
```
allmightyspiff, line 16: Add Monthly cost. The portal knows if you are building an hourly or monthly server, so it can just show you the relevant price. For the CLI we should show both, and Setup since the portal shows that too.
```suggestion
    table = formatting.KeyValueTable(['Id', 'KeyName', 'Description', 'Hourly', 'Monthly', 'Setup'])
```

SoftLayer/CLI/virt/os_available.py
```diff
@@ -0,0 +1,28 @@
...
+    table.align['name'] = 'r'
+    table.align['value'] = 'l'
+    manager = SoftLayer.OrderingManager(env.client)
+    _filter = {"items": {"prices": {"categories": {"categoryCode": {"operation": "os"}}}}}
+    operations = manager.get_items(storage_filter=_filter, package_id=46)
+
+    for operation_system in operations:
+        table.add_row([operation_system.get('id'), operation_system.get('keyName'), operation_system.get('description'),
+                       operation_system['prices'][0]['laborFee'],
```
allmightyspiff, line 25: Use the hourlyFee instead of laborFee
```suggestion
                       operation_system['prices'][0]['hourlyRecurringFee'],
                       operation_system['prices'][0]['recurringFee'],
```

allmightyspiff (changes requested):

allmightyspiff:
@caberos That analysis error is fixed in #1835. So you should just be able to merge in master to your branch and it should pass. Thanks

allmightyspiff (approved):

Issue #1844, opened 2023-01-27
unexpected error when using vlan list with --limit 0 or 999 and a large number of vlans (>400)


YuriG7:
**Describe the bug**

When I try to retrieve a large number of vlans (more than 400) the CLI returns an unexpected error. Here is the output of the command:

```
$ slcli -v vlan list --limit 0
Calling: SoftLayer_Account::getNetworkVlans(id=None, mask='firewallInterfaces,hardwareCount,primaryRouter[id, fullyQualifiedDomainName, datacenter],subnetCount,billingItem,totalPrimaryIpAddressCount,virtualGuestCount,networkSpace,networkVlanFirewall[id,fullyQualifiedDomainName,primaryIpAddress],attachedNetworkGateway[id,name,networkFirewall],tagReferences[tag[name]]', filter='{'networkVlans': {'id': {'operation': 'orderBy', 'options': [{'name': 'sort', 'value': ['ASC']}]}}}', args=(), limit=100, offset=0))
Calling: SoftLayer_Account::getNetworkVlans(id=None, mask='firewallInterfaces,hardwareCount,primaryRouter[id, fullyQualifiedDomainName, datacenter],subnetCount,billingItem,totalPrimaryIpAddressCount,virtualGuestCount,networkSpace,networkVlanFirewall[id,fullyQualifiedDomainName,primaryIpAddress],attachedNetworkGateway[id,name,networkFirewall],tagReferences[tag[name]]', filter='{'networkVlans': {'id': {'operation': 'orderBy', 'options': [{'name': 'sort', 'value': ['ASC']}]}}}', args=(), limit=100, offset=100))
Calling: SoftLayer_Account::getNetworkVlans(id=None, mask='firewallInterfaces,hardwareCount,primaryRouter[id, fullyQualifiedDomainName, datacenter],subnetCount,billingItem,totalPrimaryIpAddressCount,virtualGuestCount,networkSpace,networkVlanFirewall[id,fullyQualifiedDomainName,primaryIpAddress],attachedNetworkGateway[id,name,networkFirewall],tagReferences[tag[name]]', filter='{'networkVlans': {'id': {'operation': 'orderBy', 'options': [{'name': 'sort', 'value': ['ASC']}]}}}', args=(), limit=100, offset=200))
[...]

PR #1847, opened 2023-01-27
unexpected error when using vlan list with --limit 0 or 999 and a lar…

caberos:
[#issue1844](https://github.com/softlayer/softlayer-python/issues/1844)

image

### DIFF
--- a/SoftLayer/CLI/vlan/list.py
+++ b/SoftLayer/CLI/vlan/list.py
@@ -65,7 +65,7 @@ def cli(env, sortby, datacenter, number, name, limit):
             vlan.get('vlanNumber'),
             vlan.get('fullyQualifiedName'),
             vlan.get('name') or formatting.blank(),
-            vlan.get('networkSpace').capitalize(),
+            vlan.get('networkSpace').capitalize() or formatting.blank(),
             utils.lookup(vlan, 'primaryRouter', 'datacenter', 'name'),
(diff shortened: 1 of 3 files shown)


edsonarios (approved):
Looks good to me

SoftLayer/CLI/vlan/list.py
```diff
@@ -65,7 +65,7 @@ def cli(env, sortby, datacenter, number, name, limit):
             vlan.get('vlanNumber'),
             vlan.get('fullyQualifiedName'),
             vlan.get('name') or formatting.blank(),
-            vlan.get('networkSpace').capitalize(),
+            vlan.get('networkSpace').capitalize() or formatting.blank(),
```
allmightyspiff, line 68: Apparently Vlans with a null `networkSpace` are Direct Link vlans, so lets label them as such.

```suggestion
            vlan.get('networkSpace', 'DIRECT LINK').capitalize() ,
```

Also, I think `vlan.get('networkSpace').capitalize() or formatting.blank()` will still error on an empty `networkSpace` property. Setting a default for `vlan.get` should fix that.

allmightyspiff (changes requested):
Change the default response for a null `networkSpace` to `Direct Link`
Add a unit test for getting a vlan with an empty networkSpace property as well.

Thanks

allmightyspiff (approved):

Issue #1884, opened 2023-03-21
Dont fail when trying to get user information from `slcli event-log get`


allmightyspiff:
For example this command shouldn't fail:

```
$> slcli event-log get -l -1 -d 03/17/2023
SoftLayerAPIError(SoftLayer_Exception_PermissionDenied): You do not have permission to access this user. 
```

Basically just add a Try/Catch around https://github.com/softlayer/softlayer-python/blob/master/SoftLayer/CLI/event_log/get.py#L59
If we get that PermissionDenied exception set `username` to `Access Denied`. Add a note in the docblock for this cli command that `Access Denied` means the user you are trying to lookup isn't a child of your current user.

PR #1887, opened 2023-03-22
Dont fail when trying to get user information from slcli event-log get

caberos:
[#issue1884](https://github.com/softlayer/softlayer-python/issues/1884)

image


SoftLayer/CLI/event_log/get.py
```diff
@@ -56,8 +57,11 @@ def cli(env, date_min, date_max, obj_event, obj_id, obj_type, utc_offset, metada
         if user == "CUSTOMER":
             username = user_data.get(log['userId'])
             if username is None:
-                username = user_mgr.get_user(log['userId'], "mask[username]")['username']
-                user_data[log['userId']] = username
+                try:
+                    username = user_mgr.get_user(log['userId'], "mask[username]")['username']
+                    user_data[log['userId']] = username
+                except ValueError as ex:
+                    raise exceptions.ArgumentError("You do not have permission to access this user.  %s" % ex)
```
allmightyspiff, line 64: The point of this change is so that this command keeps going when were unable to access the user information.
```suggestion
                except exceptions.SoftLayerAPIError as ex:
                    username = log['userId']
```

allmightyspiff (changes requested):
Update the Exception as commented, and add a unit test to make sure this command has a success even if the api call to `user_mgr.get_user()` retursn a SoftLAyerAPIError

edsonarios:
Screenshots of the test result
image

Issue #1913, opened 2023-04-17
New Command: `slcli user apikey`


BrianSantivanez:
Related to: https://github.com/softlayer/softlayer-python/issues/1906
`slcli user apikey` should have 3 flags.
- --add adds a ne api key, returns the api key data
- --remove deletes the existing api key
- --refresh deletes the existing api key, and adds a new one, which is returned to the user.

Issue #1906, opened 2023-04-12
VPN password and API key management


BrianSantivanez:
## TODO

- `slcli user vpn-password` should allow a user to set the VPN password ( https://sldn.softlayer.com/reference/services/SoftLayer_User_Customer/updateVpnPassword/) for a user.
- `slcli user detail` should show if that user has an API key, and when that key was created. (dont need to display the actual api key)
- `slcli user apikey` should have 3 flags.
    - `--add` adds a ne api key, returns the api key data
    -  `--remove` deletes the existing api key
    - `--refresh` deletes the existing api key, and adds a new one, which is returned to the user

Each of these 3 main bullets should be spun off into separate issues.

PR #1919, opened 2023-04-19
New Command: `slcli user apikey`

BrianSantivanez:
Related to: https://github.com/softlayer/softlayer-python/issues/1906
Issue: https://github.com/softlayer/softlayer-python/issues/1913
Observations: the new command and its unit tests were added
image

### DIFF
--- a/SoftLayer/CLI/user/apikey.py
+++ b/SoftLayer/CLI/user/apikey.py
@@ -1,0 +1,54 @@
+"""Allows to create, remove or refresh user's API authentication key."""
+import click
+
+import SoftLayer
+from SoftLayer.CLI import environment
+from SoftLayer.CLI import exceptions
+from SoftLayer.CLI import helpers
+
+
+@click.command(cls=SoftLayer.CLI.command.SLCommand, )
+@click.argument('identifier')
+@click.option('--add', is_flag=True, default=False,
+              help="Create an user's API authentication key.")
+@click.option('--remove', is_flag=True, default=False,
+              help="Remove an user's API authentication key.")
+@click.option('--refresh', is_flag=True, default=False,
+              help="Refresh an user's API authentication key.")
+@environment.pass_env
+def cli(env, identifier, add, remove, refresh):
+
+    """Allows to create, remove or refresh user's API authentication key.
+
+    Each user can only have a single API key.
+    """
+
+    options = {
+        '--add': add,
+        '--remove': remove,
+        '--refresh': refresh,
+    }
+
+    if not any(options.values()):
+        raise exceptions.CLIAbort(f'At least one option is required: [{",".join(options.keys())}]')
+
+    if sum(options.values()) > 1:
+        raise exceptions.CLIAbort(f'Can only specify one option of [{",".join(options.keys())}]')
+
+    mgr = SoftLayer.UserManager(env.client)
+    user_id = helpers.resolve_id(mgr.resolve_ids, identifier, 'username')
+
+    if add:
+        api_authentication_key = mgr.add_api_authentication_key(user_id)
+        click.secho(f'Successfully added. New API Authentication Key: {api_authentication_key}', fg='green')
+    else:
+        api_authentication_keys = mgr.get_api_authentication_keys(user_id)
+        if len(api_authentication_keys) == 0:
+            raise exceptions.CLIAbort('The user has not API authentication keys')
+
+        mgr.remove_api_authentication_key(api_authentication_keys[0]['id'])
+        if remove:
+            click.secho('Successfully removed API authentication key', fg='green')
+        else:
+            api_authentication_key = mgr.add_api_authentication_key(user_id)
+            click.secho(f'Successfully refreshed. New API Authentication Key: {api_authentication_key}', fg='green')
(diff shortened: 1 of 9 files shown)


ramkishor-ch:
Hi,

I have "tested the changes".
Please find the screenshot for reference.

Screen Shot: 1


Thanks,
Ramkishor Chaladi.

ramkishor-ch (approved):
Hi,
I have "tested the changes".
Approving the Pull Request.
Thanks,
Ramkishor Chaladi.

SoftLayer/CLI/user/apikey.py
```diff
@@ -0,0 +1,54 @@
...
+        api_authentication_keys = mgr.get_api_authentication_keys(user_id)
+        if len(api_authentication_keys) == 0:
+            raise exceptions.CLIAbort('The user has not API authentication keys')
+
+        mgr.remove_api_authentication_key(api_authentication_keys[0]['id'])
+        if remove:
+            click.secho('Successfully removed API authentication key', fg='green')
+        else:
+            api_authentication_key = mgr.add_api_authentication_key(user_id)
+            click.secho(f'Successfully refreshed. New API Authentication Key: {api_authentication_key}', fg='green')
```
allmightyspiff, line 54: Change the flow of this section to be something like this.

```
if remove or refresh:
    # remove the api key
    # print message that a key was removed

if add or refresh:
    # add an api key
    # print message about the key being added
```
Also, if the user doesn't have an api key to remove, that shouldn't generate an error.

SoftLayer/CLI/user/apikey.py
```diff
@@ -0,0 +1,54 @@
...
+    mgr = SoftLayer.UserManager(env.client)
+    user_id = helpers.resolve_id(mgr.resolve_ids, identifier, 'username')
+
+    if add:
+        api_authentication_key = mgr.add_api_authentication_key(user_id)
+        click.secho(f'Successfully added. New API Authentication Key: {api_authentication_key}', fg='green')
+    else:
+        api_authentication_keys = mgr.get_api_authentication_keys(user_id)
+        if len(api_authentication_keys) == 0:
+            raise exceptions.CLIAbort('The user has not API authentication keys')
```
allmightyspiff, line 47: Move this logic to the manager function, and don't raise an exception on no existing api keys. If the user doesn't have any keys to remove, I consider that a success state.

allmightyspiff (changes requested):

SoftLayer/managers/user.py
```diff
@@ -332,6 +332,20 @@ def add_api_authentication_key(self, user_id):
...
         return self.user_service.addApiAuthenticationKey(id=user_id)
 
+    def get_api_authentication_keys(self, user_id):
+        """Calls SoftLayer_User_Customer::getApiAuthenticationKeys
+
+        :param int user_id: User to add API key to
+        """
+        return self.user_service.getApiAuthenticationKeys(id=user_id)
+
+    def remove_api_authentication_key(self, api_key_id):
```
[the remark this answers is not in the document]
BrianSantivanez: `remove_api_authentication_key` was updated to work with user_id instead of api_key_id and also returns true if the user does not have API key

SoftLayer/CLI/user/apikey.py
BrianSantivanez: The exception was removed and when the user does not have an API key the method returns true

SoftLayer/CLI/user/apikey.py
BrianSantivanez: The logic of command was updated following the suggestion

tests/managers/user_tests.py
```diff
@@ -334,3 +334,15 @@ def test_remove_dedicated(self):
...
+        self.manager.add_api_authentication_key(123456)
+        self.assert_called_with('SoftLayer_User_Customer', 'addApiAuthenticationKey')
+
+    def test_get_api_authentication_keys(self):
+        self.manager.get_api_authentication_keys(123456)
+        self.assert_called_with('SoftLayer_User_Customer', 'getApiAuthenticationKeys')
+
+    def test_remove_api_authentication_key(self):
+        self.manager.remove_api_authentication_key(123456)
+        self.assert_called_with('SoftLayer_User_Customer', 'removeApiAuthenticationKey')
```
allmightyspiff, line 348: Add a unit test to cover the case when a user doesn't have an API authentication key.
Make sure `SoftLayer_User_Customer::getApiAuthenticationKeys` is called, and make sure `removeApiAuthenticionKey` is called with the correct id, since it should be different from the user id you passed in.

Thanks.

allmightyspiff (changes requested):
Looks good, just add 1 unit test to cover the case of trying to remove a key from a user with 0 keys, and this will be good to go, thanks.,

tests/managers/user_tests.py
BrianSantivanez: Test when user does not have api key was added, when a user does not have an API key `removeApiAuthenticionKey ` returns True and `SoftLayer_User_Customer:removeApiAuthenticionKey` is not called, so in this test I only made sure that `SoftLayer_User_Customer::getApiAuthenticationKeys` does not return any API key

allmightyspiff (approved):

Issue #2049, opened 2023-07-19
`slcli ticket list` has imposed a `limit=100` restriction


pauljnightingale:
> Reminder: No username or APIkeys should be added to these issues, as they are public.

**Describe the bug**
`slcli ticket list` has imposed a `limit=100` restriction
This restriction is fine as a default but there is no customer option to provide an higher value, such as `slcli ticket detail N --count  N`
https://github.com/softlayer/softlayer-python/commit/ac2d7cfb8c89026d95f82feacae2c87a32591f71#diff-a938c2e33e2d945a88ce194087c36e2e39f5abfecbed0c79c984e5fab70cf5ec

**Expected behavior**
This is a change of behaviour and can result in an incomplete list being returned

Noting, this is inconsistent with `ibmcloud sl ticket list` that returns the full list 

**Screenshots**

**Version**
Include the output of `slcli --version`
```
slcli --version
Current: slcli (SoftLayer Command-line) v6.1.7
Latest:  slcli (SoftLayer Command-line) v6.1.7
```

allmightyspiff:
@JayasilanJibm Can you take a look at this one. thanks

PR #2072, opened 2023-08-08
slcli ticket list command with user limit option #2049

JayasilanJibm:
Hi Chris,

Please review this PR.
Fixes: #2049

Changes:
 - slcli ticket list command with user limit option.

Screenshot 2023-08-08 at 5 35 49 PM

Thanks,
Jayasilan

### DIFF
--- a/SoftLayer/CLI/ticket/list.py
+++ b/SoftLayer/CLI/ticket/list.py
@@ -13,20 +13,17 @@ from SoftLayer.CLI import formatting
 @click.option('--open / --closed', 'is_open', default=True,
               help="Display only open or closed tickets")
 @click.option('--limit', default=100, show_default=True, type=click.INT, help="Result limit")
-@click.option("--all", "-a", "all_open", is_flag=True, default=False, help="Results all Open tickets")
+@click.option("--all", "-a", "all_tickets", is_flag=True, default=False, help="Return all tickets")
 @environment.pass_env
-def cli(env, is_open, limit, all_open):
+def cli(env, is_open, limit, all_tickets):
     """List tickets."""
     ticket_mgr = SoftLayer.TicketManager(env.client)
     table = formatting.Table([
         'id', 'Case_Number', 'assigned_user', 'title', 'last_edited', 'status', 'updates', 'priority'
     ])
 
-    if all_open and is_open:
-        tickets = ticket_mgr.list_tickets(open_status=is_open, closed_status=not is_open, limit=limit,
-                                          all_tickets=all_open)
-    else:
-        tickets = ticket_mgr.list_tickets(open_status=is_open, closed_status=not is_open, limit=limit)
+    tickets = ticket_mgr.list_tickets(open_status=is_open,
+                                      closed_status=not is_open, limit=limit, all_tickets=all_tickets)
 
     for ticket in tickets:
         user = formatting.blank()
--- a/SoftLayer/CLI/ticket/list.py
+++ b/SoftLayer/CLI/ticket/list.py
@@ -12,15 +12,16 @@ from SoftLayer.CLI import formatting
 @click.command(cls=SoftLayer.CLI.command.SLCommand, )
 @click.option('--open / --closed', 'is_open', default=True,
               help="Display only open or closed tickets")
+@click.option('--limit', default=100, show_default=True, type=click.INT, help="Result limit")
 @environment.pass_env
-def cli(env, is_open):
+def cli(env, is_open, limit):
     """List tickets."""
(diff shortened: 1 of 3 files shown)


ramkishor-ch (approved):

allmightyspiff (changes requested):
The changes you have are good, but I would also like to add a `--all` option.

If a user specifies `slcli ticket list --all` it would get all the open tickets, regardless of the limit. 

Something like this in the manager
` def list_tickets(self, open_status=True, closed_status=True, limit=None, all_tickets=False):`
`return self.client.call('Account', call, mask=mask, iter=all_tickets, limit=limit)`

Also add a unit test for these 2 new flags (`limit` and `all`)
thanks

SoftLayer/managers/ticket.py
```diff
@@ -22,24 +22,30 @@ def __init__(self, client):
         self.account = self.client['Account']
         self.ticket = self.client['Ticket']
 
-    def list_tickets(self, open_status=True, closed_status=True):
+    def list_tickets(self, open_status=True, closed_status=True, limit=None, all_tickets=None):
```
allmightyspiff, line 25: Set the default for `all_tickets` to just be False, so you don't have to use the `if` statement later in this function.
```suggestion
    def list_tickets(self, open_status=True, closed_status=True, limit=None, all_tickets=False):
```

SoftLayer/managers/ticket.py
```diff
@@ -22,24 +22,30 @@ def __init__(self, client):
...
 
         :param boolean open_status: include open tickets
         :param boolean closed_status: include closed tickets
+        :param int limit: result limit
         """
         mask = """mask[id, serviceProviderResourceId, title, assignedUser[firstName, lastName], priority,
                   createDate, lastEditDate, accountId, status, updateCount]"""
 
         call = 'getTickets'
+        # print('All Checks ',open_status, closed_status,all_tickets)
```
allmightyspiff, line 36: Remove this line

SoftLayer/managers/ticket.py
```diff
@@ -22,24 +22,30 @@ def __init__(self, client):
...
             elif closed_status:
                 call = 'getClosedTickets'
             else:
                 raise ValueError("open_status and closed_status cannot both be False")
-        return self.client.call('Account', call, mask=mask, iter=False, limit=100)
+
+        if all([open_status, all_tickets]):
+            return self.client.call('Account', call, mask=mask, iter=all_tickets, limit=limit)
+
+        return self.client.call('Account', call, mask=mask, iter=False, limit=limit)
```
allmightyspiff, line 48: You don't need this if statement if we set the default for `all_tickets` to False.  We also want to allow users the options to get all closed tickets as well, and this change will allow that.

```suggestion
        return self.client.call('Account', call, mask=mask, iter=all_tickets, limit=limit)
```

SoftLayer/CLI/ticket/list.py
```diff
@@ -12,15 +12,22 @@
...
         'id', 'Case_Number', 'assigned_user', 'title', 'last_edited', 'status', 'updates', 'priority'
     ])
 
-    tickets = ticket_mgr.list_tickets(open_status=is_open, closed_status=not is_open)
+    if all_open and is_open:
+        tickets = ticket_mgr.list_tickets(open_status=is_open, closed_status=not is_open, limit=limit,
+                                          all_tickets=all_open)
+    else:
+        tickets = ticket_mgr.list_tickets(open_status=is_open, closed_status=not is_open, limit=limit)
+
```
allmightyspiff, line 30: This can be simplified to the following:
```suggestion
         tickets = ticket_mgr.list_tickets(open_status=is_open, closed_status=not is_open, limit=limit, all_tickets=all_tickets)

```

SoftLayer/CLI/ticket/list.py
```diff
@@ -12,15 +12,22 @@
 @click.command(cls=SoftLayer.CLI.command.SLCommand, )
 @click.option('--open / --closed', 'is_open', default=True,
               help="Display only open or closed tickets")
+@click.option('--limit', default=100, show_default=True, type=click.INT, help="Result limit")
+@click.option("--all", "-a", "all_open", is_flag=True, default=False, help="Results all Open tickets")
```
allmightyspiff, line 16:
```suggestion
@click.option("--all", "-a", "all_tickets", is_flag=True, default=False, help="Returns all tickets")
```

allmightyspiff (changes requested):
The `--all` option should return all tickets, open or closed. I apologize if that wasn't clear in the issue description.

allmightyspiff (approved):

PR #2115, opened 2023-11-16
GH Action for snap

kz6fittycent:
Builds and performs checks against the new snap after commit.

### DIFF
--- a/.github/workflows/test-snap-can-build.yml
+++ b/.github/workflows/test-snap-can-build.yml
@@ -1,0 +1,25 @@
+name: 🧪 Test snap can be built on x86_64
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+    
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: snapcore/action-build@v1
+        id: build
+
+      - uses: diddlesnaps/snapcraft-review-action@v1
+        with:
+          snap: ${{ steps.build.outputs.snap }}
+          isClassic: 'false'
+          # Plugs and Slots declarations to override default denial (requires store assertion to publish)
+          # plugs: ./plug-declaration.json
+          # slots: ./slot-declaration.json


allmightyspiff:
Currently I have snapcraft setup to run these builds, and they seem to be working. Although they are only getting published to the Edge section.  I manually promoted v6.1.10 from Edge to Stable for now.

What would be helpful is a GH action that runs on release that automatically builds and deploys to Stable any release, that is something I had a really hard time getting to work last time I tried to solve this problem, and I guess I just gave up on it.

Also, this Action already has a warning, so fix that first:
```
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, diddlesnaps/snapcraft-review-action@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
```
I won't merge an Action that has outdated libraries. I think this is why I stopped building snaps with GH in the first place and moved it to snapcraft itself for the builds.

This check should build against the other architectures we support (everything except i386, I forget why I'm not building this one anymore, but I think it has dependency issues or something) to be a valid test.

## TODO
- Fix the warning. 
- Build all architectures
- Change to run on release instead of master
- If possible, publish the new build to Stable. 

Anyway, let me know if there are any issues with the above requests, I don't really know a lot about the snap build process. Or if there is anything I can do to make the releases more automatic. 
Thanks for the pull request :)

allmightyspiff (changes requested):
See https://github.com/softlayer/softlayer-python/pull/2115#issuecomment-1815042622

kz6fittycent:
> Currently I have snapcraft setup to run these builds, and they seem to be working. Although they are only getting published to the Edge section. I manually promoted v6.1.10 from Edge to Stable for now.
> 
> What would be helpful is a GH action that runs on release that automatically builds and deploys to Stable any release, that is something I had a really hard time getting to work last time I tried to solve this problem, and I guess I just gave up on it.
> 
> Also, this Action already has a warning, so fix that first:
> 
> ```
> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2, diddlesnaps/snapcraft-review-action@v1. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
> ```
> 
> I won't merge an Action that has outdated libraries. I think this is why I stopped building snaps with GH in the first place and moved it to snapcraft itself for the builds.
> 
> This check should build against the other architectures we support (everything except i386, I forget why I'm not building this one anymore, but I think it has dependency issues or something) to be a valid test.
> 
> ## TODO
> * Fix the warning.
> * Build all architectures
> * Change to run on release instead of master
> * If possible, publish the new build to Stable.
> 
> Anyway, let me know if there are any issues with the above requests, I don't really know a lot about the snap build process. Or if there is anything I can do to make the releases more automatic. Thanks for the pull request :)

That's fine. I'll close the PR, as it's not totally necessary. You're right about the warning, and that's something I'll have to check into, since I didn't create the action - I always considered it a non-issue. 

That said, the good folks at Canonical are working on a GH action that can publish to the channel of choice - it's still in testing but once done, I'll see if I can muster the time to send you another PR.

PR #2222, opened 2025-05-22
Support Click 8.2+

s-t-e-v-e-n-k:
Click 8.2 and above will now force an abort if a confirmation prompt isn't answered, rather than raising the CLIAbort that is expected. Catch this exception so that our own exceptions are raised.


s-t-e-v-e-n-k:
Done!

PR #2230, opened 2025-07-16
Remove unused dependency prettytable

karyotakisg:
As part of our ongoing research on Python dependency management we noticed a potential improvement in your project’s dependency management.

Specifically, the dependency `prettytable` is specified as a requirement in the configuration files, when in reality it is not needed.

The dependency is unused since 1238377, where its import was removed.

Hope this is helpful!

Best regards

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -173,7 +173,6 @@ If you cannot install python 3.6+ for some reason, you will need to use a versio
 
 Python Packages
 ---------------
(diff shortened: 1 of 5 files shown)


allmightyspiff (approved):

allmightyspiff:
Thanks for the commit, looks good.

